PR #2, opened 2021-09-17
Add skeleton dropwizard app

gidsg:
Doesn't do much but should start up, render page at `/` and have healthcheck.


pom.xml
```diff
@@ -0,0 +1,118 @@
...
+        </dependency>
+        <dependency>
+            <groupId>jakarta.validation</groupId>
+            <artifactId>jakarta.validation-api</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.hibernate.validator</groupId>
+            <artifactId>hibernate-validator</artifactId>
+        </dependency>
+    </dependencies>
```
alexbishop1, line 52: I think we want to set explicit versions for these so we can control the updates.

pom.xml
```diff
@@ -0,0 +1,118 @@
...
+
+    <build>
+        <plugins>
+            <plugin>
+                <artifactId>maven-shade-plugin</artifactId>
+                <version>3.2.4</version>
+                <configuration>
```
alexbishop1, line 59: In other POMs we have `<outputFile>target/${project.artifactId}-${project.version}-allinone.jar</outputFile>` — do we want something similar here to be consistent? We have to reference the file name in the Dockerfile and possibly other places.

pom.xml
```diff
@@ -0,0 +1,118 @@
...
+                <artifactId>maven-surefire-plugin</artifactId>
+                <version>3.0.0-M5</version>
+            </plugin>
+        </plugins>
+    </build>
+
+    <reporting>
+        <plugins>
+        </plugins>
+    </reporting>
```
alexbishop1, line 117: Do we need this section if it’s empty?

src/main/java/uk/gov/pay/webhooks/WebhooksApp.java
```diff
@@ -0,0 +1,27 @@
...
+
+public class WebhooksApp extends Application<WebhooksConfiguration> {
+    public static void main(String[] args) throws Exception {
+        new WebhooksApp().run(args);
+    }
+
+@Override
+public void run(WebhooksConfiguration configuration,
+                Environment environment) {
+    System.out.println("hello world");
```
alexbishop1, line 17: Not sure we want to leave this in.

src/main/java/uk/gov/pay/webhooks/healthcheck/HealthCheckResource.java
```diff
@@ -0,0 +1,30 @@
...
+    @Inject
+    public HealthCheckResource(Environment environment) {
+        this.environment = environment;
+    }
+
+    @GET
+    @Path("/")
+    @Produces(APPLICATION_JSON)
+    public Response healthCheck() {
+        var ok = ImmutableMap.of("healthy", "true");
```
alexbishop1, line 27: We can use `Map.of("healthy", "true")` here.

src/main/resources/banner.txt
```diff
@@ -0,0 +1,10 @@
+       ____
+      / __ \____ ___  __
+     / /_/ / __ `/ / / /
+    / ____/ /_/ / /_/ /
+   /_/    \__,_/\__, /
+               /____/
```
alexbishop1, line 6: Snazzy!

alexbishop1 (changes requested):
Looks like basically what we want.

src/main/java/uk/gov/pay/webhooks/WebhooksApp.java
gidsg: removed

src/main/java/uk/gov/pay/webhooks/healthcheck/HealthCheckResource.java
gidsg: done

gidsg:
@alexbishop1 I have force-pushed to address your comments and made a couple of other tweaks in separate comments.

pom.xml
gidsg: done

pom.xml
gidsg: have added explicit versions and removed the ones we aren't currently using.

alexbishop1 (approved):
This will do for now.

PR #3, opened 2021-09-21
PP-8443 Configure database setup

sfount:
Adds dependencies required to setup, configure and run migrations on a Postgres database: 
* dropwizard core migrations
* postgres driver

Sets up dependency injection in line with other apps (using Guice) to set the standard and clean up existing healthcheck endpoint. 

Adds first database migration to setup a simple table.

Adds standard commons database healtcheck with assertion on healthy database connection.

Sets up test environment to run tests against an actual database
* JUnit5
* testcontainers
* hamcreset matchers
* reset assured

### DIFF
--- a/src/main/resources/migrations/0001_create_table_webhooks.sql
+++ b/src/main/resources/migrations/0001_create_table_webhooks.sql
@@ -1,0 +1,18 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:create-table-webhooks
+
+CREATE TYPE live_or_test AS ENUM('LIVE', 'TEST');
+
+CREATE table webhooks (
+    id SERIAL PRIMARY KEY,
+    created_at TIMESTAMP WITHOUT TIME ZONE NOT NULL,
+    external_id VARCHAR(30) NOT NULL,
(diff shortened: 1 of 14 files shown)


src/main/resources/migrations/0001_create_table_webhooks.sql
```diff
@@ -0,0 +1,18 @@
...
+--changeset uk.gov.pay:create-table-webhooks
+
+CREATE TYPE live_or_test AS ENUM('LIVE', 'TEST');
+
+CREATE table webhooks (
+    id SERIAL PRIMARY KEY,
+    created_at TIMESTAMP WITHOUT TIME ZONE NOT NULL,
+    external_id VARCHAR(30) NOT NULL,
+    service_id VARCHAR(30) NOT NULL,
+    type live_or_test NOT NULL,
```
mrlumbu, line 12: any reason we went enum over boolean here (given we went the other way in connector and ledger)? Are we going to hate ourselves when we find ourselves writing enum deserialisers all over the shop in a weeks time?

mrlumbu (dismissed):
All looks sensible. A question for your consideration

mrlumbu (approved):

PR #4, opened 2021-09-23
PP-8443 Make mvn verify work

mrlumbu:
Who knows why this works. Who cares

PR #5, opened 2021-09-23
BAU GitHub Action for integration tests

sfount:
Depends on https://github.com/alphagov/pay-webhooks/pull/4

Use JDK 17 as specified in the pom file.

Run on standard Ubuntu base, similar to unit tests run by other repos.

Output of action should be the result of `mvn clean integration-test`.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -1,0 +1,31 @@
+name: Github Actions Tests
+
+on:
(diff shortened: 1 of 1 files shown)


mrlumbu:
why not `maven clean verify`? (otherwise post-integration-test cleanup step isn't run)

mrlumbu (approved):

PR #6, opened 2021-09-24
PP-8443 Switch from EclipseLink to Hibernate

alexbishop1:
Dropwizard default JPA implementation is Hibernate.

We were having problems using EL with Java 17 in Dropwizard. Hibernate works quite straightforwardly - `WebhookResourceIT` is a passing test using new Hibernate things.

with @mrlumbu, @sfount

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -23,6 +23,8 @@
         <testcontainers.version>1.16.0</testcontainers.version>
         <pay-java-commons.version>1.0.20210914101116</pay-java-commons.version>
         <mainClass>uk.gov.pay.webhooks.app.WebhooksApp</mainClass>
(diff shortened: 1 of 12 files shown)


pom.xml
```diff
@@ -85,6 +97,11 @@
             <artifactId>utils</artifactId>
             <version>${pay-java-commons.version}</version>
         </dependency>
+        <dependency>
+            <groupId>uk.gov.service.payments</groupId>
+            <artifactId>model</artifactId>
```
sfount, line 102: looks like this is no longer used by the PR but I assume it will be used later down the line to have standard responses from backends

sfount (approved):

PR #7, opened 2021-09-24
PP-8475 Dockerfile

mrlumbu:
Dockerfile added same as other java apps.

Have used standard docker startup script, but commented out
wait on dependencies bit until we have some dependencies.

Moved config to the right place so that it is included
in target.

Added dependabot file to keep base image up to date.

```
mvn clean package
docker build . -t webhooks
docker run -e JAVA_OPTS='-Xms128m -Xmx1G' -it webhooks
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,29 @@
+FROM openjdk:17-jdk-alpine@sha256:a996cdcc040704ec6badaf5fecf1e144c096e00231a29188596c784bcf858d05
+
+RUN ["apk", "--no-cache", "upgrade"]
+
+ARG DNS_TTL=15
+
+# Default to UTF-8 file.encoding
+ENV LANG C.UTF-8
+
+RUN echo networkaddress.cache.ttl=$DNS_TTL >> "$JAVA_HOME/conf/security/java.security"
+
+# Add RDS CA certificates to the default truststore
+RUN wget -qO - https://s3.amazonaws.com/rds-downloads/rds-ca-2019-root.pem       | keytool -importcert -noprompt -cacerts -storepass changeit -alias rds-ca-2019-root \
+ && wget -qO - https://s3.amazonaws.com/rds-downloads/rds-combined-ca-bundle.pem | keytool -importcert -noprompt -cacerts -storepass changeit -alias rds-combined-ca-bundle
+
+RUN ["apk", "add", "--no-cache", "bash", "tini"]
+
+ENV PORT 8080
+EXPOSE 8080
+
+WORKDIR /app
+
+COPY docker-startup.sh /app/docker-startup.sh
+COPY target/*.yaml /app/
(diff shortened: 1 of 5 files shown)


katstevens:
Swinging by to suggest enabling Dependabot updates for this base image, if this isn't on your radar already (so it doesn't get forgotten later).  [Something like this example](https://github.com/alphagov/pay-endtoend/blob/master/.github/dependabot.yml#L13)?

Dockerfile
```diff
@@ -0,0 +1,29 @@
+FROM openjdk:17-jdk-alpine@sha256:a996cdcc040704ec6badaf5fecf1e144c096e00231a29188596c784bcf858d05
```
alexbishop1, line 1: This is not quite the same as we use in our other Java apps (apart from the obvious Java 17 v Java 11 difference).

We currently use [Docker images provided by the Adoptium (formerly AdoptOpenJDK) project](https://hub.docker.com/u/adoptopenjdk) that use [Alpine Linux](https://www.alpinelinux.org/) as the operating system. Since Alpine works with [musl](https://musl.libc.org/) rather than [glibc](https://www.gnu.org/software/libc/), they toss in the [alpine-pkg-glibc](https://github.com/sgerrand/alpine-pkg-glibc) glibc compatibility layer and then include the standard glibc-based Linux build of the Eclipse Temurin (formerly AdoptOpenJDK) JRE.

I should note that while our Java apps have used this configuration for a long time, the Alpine Linux maintainers would disagree. They consider the glibc compatibility layer to be so dangerous that they’ve considered [making changes to Alpine Linux to stop it from working with alpine-pkg-glibc](https://github.com/docker-library/official-images/issues/10794). Their disapproval is having effects: [Adoptium will not be making Alpine images for Java 17 that use alpine-pkg-glibc](https://github.com/adoptium/containers/issues/1) and [there will not be official Alpine OpenJDK images that use alpine-pkg-glibc](https://github.com/docker-library/official-images/pull/10779) either.

What this PR uses is an [official OpenJDK Docker image](https://hub.docker.com/_/openjdk) that uses Alpine Linux as the operating system and a native musl-based Alpine Linux port of OpenJDK (developed under the name [Project Portola](https://openjdk.java.net/projects/portola/)) (not a JRE). The Alpine Linux OpenJDK port is not currently considered to be production ready.

That said, I think it’s fine to use for now. We can always re-evaluate later.

There’s also the JDK versus JRE difference. Since Java 11, the JRE has not really been a thing but the AdoptOpenJDK project continued to produce JRE binaries and images. Adoptium will only produce [Temurin JDK (and not JRE) builds for Java 17 and above](https://github.com/adoptium/temurin-build/issues/2683)) so at some point we’re going to have to deal with this change (and maybe look into the suggested alternative of using jlink).

alexbishop1 (approved):

PR #8, opened 2021-09-28
PP-8443 Add Jackson validation to webhook resource

mrlumbu:
Rename generic DTO to describe it as the serialiser for creating webhook
requests.

Build webhook entities from the create webhook request.

Add Jackson/ Javax annotations to the serialiser and resource to ensure
the required values are set, and are within correct constraints.

Unit test the resource layer for validations.

### DIFF
--- a/src/test/java/uk/gov/pay/webhooks/webhook/WebhookResourceTest.java
+++ b/src/test/java/uk/gov/pay/webhooks/webhook/WebhookResourceTest.java
@@ -1,0 +1,75 @@
+package uk.gov.pay.webhooks.webhook;
+
+import io.dropwizard.testing.junit5.DropwizardExtensionsSupport;
+import io.dropwizard.testing.junit5.ResourceExtension;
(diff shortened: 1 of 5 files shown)


src/test/java/uk/gov/pay/webhooks/webhook/WebhookResourceTest.java
```diff
@@ -0,0 +1,75 @@
...
+                    "callback_url": "https://some-callback-url.com",
+                    "live": true
+                }
+                """;
+        var response = resources
+                .target("/v1/webhook")
+                .request()
+                .post(Entity.json(json));
+
+        assertThat(response.getStatus(), is(422));
```
sfount, line 73: In the future we probably want to assert on the on the actual reason the validation is being rejected to make sure tests don't slip and pass for the wrong reason. This will likely go through a fair bit of iteration so I think this is fine for now

sfount (approved):

PR #10, opened 2021-09-29
PP-8506 Get Webhook resource

mrlumbu:
Adds ability to get webhook by external id.

Controversial things what I have done:

- used records for webhook create request and
webhook response. This seems pretty nice:
clearer, less verbose, more immutable.

- reorganised a bit, so dao and resource are
two packages under webhook. Think this makes sense.

- used named typed query and defined this on
Webhook entity. To my eye this makes everything
a bit cleaner and I like that queries are defined
with entity, and then DAO is responsible for
executing and handling result. Open to discuss this.

- removed mocking of Webhook thing from resource test:
this was breaking Jackson serialisation, and seems
unnecessary

- written one big WebhookResourceIT that posts to API
then retrieves from API. Seems like a nice test to have for now
and avoids need to squirt things in to DB to set up. May not
be sustainable.


src/main/java/uk/gov/pay/webhooks/webhook/resource/CreateWebhookRequest.java
```diff
@@ -0,0 +1,14 @@
...
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import javax.validation.constraints.NotEmpty;
+import javax.validation.constraints.NotNull;
+import javax.validation.constraints.Size;
+
+public record CreateWebhookRequest(
+        @JsonProperty("service_id") @NotEmpty @Size(max = 30) String serviceId,
+        @JsonProperty("live") @NotNull Boolean live,
```
gidsg, line 11: should this be `boolean` i.e. do we want to allow null values?

src/main/java/uk/gov/pay/webhooks/webhook/resource/WebhookResponse.java
```diff
@@ -0,0 +1,27 @@
+package uk.gov.pay.webhooks.webhook.resource;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import uk.gov.pay.webhooks.webhook.dao.entity.WebhookEntity;
+import uk.gov.pay.webhooks.webhook.dao.entity.WebhookStatus;
+
+public record WebhookResponse(
+        @JsonProperty("service_id") String serviceId,
+        @JsonProperty("live") Boolean live,
```
gidsg, line 9: same question as https://github.com/alphagov/pay-webhooks/pull/10/files#r718368023

src/main/java/uk/gov/pay/webhooks/webhook/resource/CreateWebhookRequest.java
sfount: AFAIK you can't annotate primitive boolean because even though you don't want to allow null values - values _can_ be null

src/main/java/uk/gov/pay/webhooks/webhook/dao/entity/WebhookEntity.java
```diff
@@ -1,60 +1,98 @@
...
 import javax.persistence.Table;
 import java.math.BigInteger;
 import java.security.SecureRandom;
 import java.time.Instant;
 import java.util.Date;
 
+ 
+@NamedQuery(
+    name = WebhookEntity.GET_BY_EXTERNAL_ID,
+    query = "select p from WebhookEntity p where externalId = :externalId"
```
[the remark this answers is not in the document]
mrlumbu: it's just the alias for the JPA entity so can be anything. So yes, should probably just call it webhook. Can do in follow up PR if that's ok

gidsg (approved):
LGTM 👍  comment can be addressed in subsequent PR

PR #11, opened 2021-09-29
PP-8443 add subscription persistence

mrlumbu:
Adds migrations and JPA things for event subscriptions.

As agreed we will use a join table to map webhooks to event
types. Ihave modelled inJPA as one to many because I don't think
we will ever care about tracing back from one event type
to see all the webhooks subscribed.
The migration includes the creation of a single event type
(card_payment_captured). This is largely to get things off
the ground and am happy to rename etc.
I have used an enum to model the name of the event type,
and then used some slightly fancy @transient logic to
automatically go from Enum to string and back. This means
we can use nice type safe enum through the business logic
and have it serialised to db correctly.
I have used a Dropwixard DAO test to test persistence layer
logic. I will add full integration test coverage
once resource end of things is done.

### DIFF
--- a/src/main/java/uk/gov/pay/webhooks/eventtype/EventTypeName.java
+++ b/src/main/java/uk/gov/pay/webhooks/eventtype/EventTypeName.java
@@ -1,0 +1,24 @@
+package uk.gov.pay.webhooks.eventtype;
+
+import java.util.stream.Stream;
+
+public enum EventTypeName {
+    CARD_PAYMENT_CAPTURED("card_payment_captured");
+    
+    private final String name;
+    
+    EventTypeName(String name) {
+        this.name = name;
+    }
+
+    public static EventTypeName of(String name) {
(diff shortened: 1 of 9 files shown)


src/main/java/uk/gov/pay/webhooks/eventtype/EventTypeName.java
```diff
@@ -0,0 +1,24 @@
...
+    
+    private final String name;
+    
+    EventTypeName(String name) {
+        this.name = name;
+    }
+
+    public static EventTypeName of(String name) {
+        return Stream.of(EventTypeName.values())
+                .filter(p -> p.getName().equals(name))
```
gidsg, line 16: could we call this `e` or similar to reference what we are filtering?

gidsg (approved):
Don't know much about the hibernate side of things but otherwise this looks good to me 👍

PR #12, opened 2021-09-29
PP-8443 add subscriptions to webhook resource

mrlumbu:
Adds functionality to set webhook subscriptions on
webhook creation, and retrieve subscriptions as part of
webhook.

I have used a webhook service to encapsulate the small
amount of business logic needed to deal with subscriptions.

There is probably more to do on validation, but using
an enum for event type names gets us a lot of the way
there.

### DIFF
--- a/src/main/java/uk/gov/pay/webhooks/eventtype/EventTypeName.java
+++ b/src/main/java/uk/gov/pay/webhooks/eventtype/EventTypeName.java
@@ -1,8 +1,11 @@
 package uk.gov.pay.webhooks.eventtype;
 
+import com.fasterxml.jackson.annotation.JsonProperty;
+
 import java.util.stream.Stream;
 
 public enum EventTypeName {
+    @JsonProperty("card_payment_captured")
     CARD_PAYMENT_CAPTURED("card_payment_captured");
     
     private final String name;
@@ -13,7 +16,7 @@ public enum EventTypeName {
 
     public static EventTypeName of(String name) {
         return Stream.of(EventTypeName.values())
-                .filter(p -> p.getName().equals(name))
+                .filter(n -> n.getName().equals(name))
                 .findFirst()
                 .orElseThrow(IllegalArgumentException::new);
     }
--- a/src/main/java/uk/gov/pay/webhooks/webhook/WebhookService.java
+++ b/src/main/java/uk/gov/pay/webhooks/webhook/WebhookService.java
@@ -1,0 +1,42 @@
+package uk.gov.pay.webhooks.webhook;
+
+import uk.gov.pay.webhooks.eventtype.EventTypeName;
+import uk.gov.pay.webhooks.eventtype.dao.EventTypeDao;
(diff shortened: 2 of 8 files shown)


sfount (approved):
Record proving to be a really clean abstraction, service swapped in at the right amount of complexity, well covered by tests 💯

src/main/java/uk/gov/pay/webhooks/webhook/WebhookService.java
```diff
@@ -0,0 +1,42 @@
...
+    public WebhookService(WebhookDao webhookDao, EventTypeDao eventTypeDao) {
+        this.webhookDao = webhookDao;
+        this.eventTypeDao = eventTypeDao;
+    }
+
+    public WebhookEntity createWebhook(CreateWebhookRequest createWebhookRequest) {
+        var webhookEntity = WebhookEntity.from(createWebhookRequest);
+        if (createWebhookRequest.subscriptions() != null) {
+            List<EventTypeEntity> subscribedEventTypes = createWebhookRequest.subscriptions().stream()
+                    .map(eventTypeName -> eventTypeDao.findByName(eventTypeName))
```
sfount, line 29: We might want to avoid a database query in a map in the future but I think given how rarely this will be called and the theoretical limit to the number of entries in the table this shouldn't an issue

src/main/java/uk/gov/pay/webhooks/eventtype/EventTypeName.java
```diff
@@ -1,8 +1,11 @@
 package uk.gov.pay.webhooks.eventtype;
 
+import com.fasterxml.jackson.annotation.JsonProperty;
+
 import java.util.stream.Stream;
 
 public enum EventTypeName {
+    @JsonProperty("card_payment_captured")
     CARD_PAYMENT_CAPTURED("card_payment_captured");
```
sfount, line 9: Relating to the validation on this PR, it feels a bit fiddly to have an enum with values that are completely unrelated to the underlying database values 
https://github.com/alphagov/pay-webhooks/blob/main/src/main/resources/migrations/0002_create_table_event_types.sql#L10

As the validation will only happen occasionally would it be reasonable to just lookup possible values from the database instead? (These might then also be used to associate the filtered list with the webhook entity)

This might be more complex if the enum is relied on by other parts of the code too.

mrlumbu: Hmm maybe. I quite like the fact that this enum (which is used all the way down from resource to dao) represents the fact that this can only have a limited set of values. e.g. I actually had to delete a test when I introduced it because I could no longer create a certain failure condition. But I agree that the disconnect with the source of truth is a bit annoying. I guess the true way is to have it as an enum in db, although that still wouldn't guarantee sync. When you say the validation will happen occasionally you mean 'every time a webhook is created'. Which isn't 1000s of tx per sec, but is not (or may not be) negligible either. Worth thinking about further

PR #13, opened 2021-09-30
Pp 8475 build and push

mrlumbu:


katstevens:
Might be a bit of a faff to set up, but have we tested that a workflow action on a different repo can't assume the role?

.github/workflows/run-tests.yml
```diff
@@ -27,5 +26,49 @@ jobs:
...
+    needs: build
+    runs-on: ubuntu-latest
+    permissions:
+      id-token: write
+      contents: read
+    steps:
+      - run: sleep 5
+      - name: Fake push
+        run: |
+          export AWS_ROLE_ARN=arn:aws:iam::223851549868:role/github_actions_role
```
sfount, line 63: It might be overkill but I wonder if we include the ARN as a GitHub secret, that way if it does get out there somehow we're not fussed but otherwise it's only available to project contributors.

All of the linked posts and references just include it directly and a brief search doesn't bring up a clear best practice for keeping them hidden (other than favouring templating them in from the thing provisioning them which doesn't apply in this case)

mrlumbu: Yeah probably a good plan. A bit security by obscurity, but given it's not really any  effort might as well.

mrlumbu:
@katstevens evidence: https://github.com/alphagov/pay-dockerfiles/pull/30/checks?check_run_id=3792668264

PR #23, opened 2021-10-28
PP-8496: add PATCH webhook endpoint

gidsg:
Add json patch endpoint at `PATCH /v1/webhook/{externalId}` which supports replace commands for updating fields.

### DIFF
--- a/src/main/java/uk/gov/pay/webhooks/webhook/resource/WebhookResource.java
+++ b/src/main/java/uk/gov/pay/webhooks/webhook/resource/WebhookResource.java
@@ -1,8 +1,12 @@
 package uk.gov.pay.webhooks.webhook.resource;
 
+import com.fasterxml.jackson.databind.JsonNode;
 import io.dropwizard.hibernate.UnitOfWork;
+import uk.gov.pay.webhooks.validations.WebhookRequestValidator;
 import uk.gov.pay.webhooks.webhook.WebhookService;
 import uk.gov.pay.webhooks.webhook.dao.entity.WebhookEntity;
+import uk.gov.service.payments.commons.api.exception.ValidationException;
+import uk.gov.service.payments.commons.model.jsonpatch.JsonPatchRequest;
 
 import javax.inject.Inject;
 import javax.validation.Valid;
@@ -11,6 +15,7 @@ import javax.ws.rs.BadRequestException;
 import javax.ws.rs.Consumes;
 import javax.ws.rs.GET;
 import javax.ws.rs.NotFoundException;
+import javax.ws.rs.PATCH;
 import javax.ws.rs.POST;
 import javax.ws.rs.Path;
 import javax.ws.rs.PathParam;
@@ -18,6 +23,7 @@ import javax.ws.rs.Produces;
 import javax.ws.rs.QueryParam;
 
 import java.util.List;
+import java.util.stream.StreamSupport;
 
 import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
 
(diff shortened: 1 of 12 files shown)


src/test/java/uk/gov/pay/webhooks/webhook/resource/WebhookResourceIT.java
```diff
@@ -52,7 +52,7 @@ public void shouldCreateAndRetrieveAWebhook() {
                 .body("live", is(true))
                 .body("callback_url", is("https://example.com"))
                 .body("description", is("description"))
-                .body("status", is("ACTIVE"))
+                .body("status", is("active"))
```
sfount, line 55: Is this desired behaviour? Any consumers of this will need to be updated (I know there happens to be one in the admin tool, it could easily lower whatever it gets from the backend so its not case sensitive)

sfount: If possible I think I'd recommend not changing the external facing statuses to keep them caps -- I'm sure there are inconsistencies with how other enums are presented on our internal endpoints so don't know if there's any exact truth to point to.

sfount (approved):
A number of questions, code looks good and scenarios well covered 👍

src/test/java/uk/gov/pay/webhooks/webhook/resource/WebhookResourceIT.java
gidsg: @sfount wasn't sure about this either 🤷‍♂️ I did this to match what we do with the event_types e.g. https://github.com/alphagov/pay-webhooks/blob/610fd677c5dd8f7925c9258741e9d4d5b9449dd6/src/main/java/uk/gov/pay/webhooks/eventtype/EventTypeName.java#L9 that said if people prefer all caps for this, that should be simpler and could remove some code.

src/main/java/uk/gov/pay/webhooks/webhook/resource/WebhookResource.java
```diff
@@ -27,10 +33,13 @@
 public class WebhookResource {
     
     private final WebhookService webhookService;
+    private final WebhookRequestValidator requestValidator;
     
     @Inject
-    public WebhookResource(WebhookService webhookService) {
+    public WebhookResource(WebhookService webhookService,
+                           WebhookRequestValidator requestValidator) {
```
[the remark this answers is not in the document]
gidsg: good question, I did this as I based the approach initially on what is done in pay-products e.g. in https://github.com/alphagov/pay-products/blob/63933cadecb98a4a4a2f04bd48d3bd63699b4946/src/main/java/uk/gov/pay/products/resources/ProductResource.java#L51 that said I can try doing this without injection and see if there are any issues.

src/test/java/uk/gov/pay/webhooks/webhook/resource/WebhookResourceIT.java
gidsg: @sfount OK I restored the previous behaviour in f1fd2f8 (done as separate commit so easier to change mind if needed)

src/main/java/uk/gov/pay/webhooks/webhook/resource/WebhookResource.java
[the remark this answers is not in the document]
sfount: Understood 👍 might be good to get people's general perspective on this but I don't think it should be blocking anything

sfount (approved):

src/main/java/uk/gov/pay/webhooks/webhook/resource/WebhookResource.java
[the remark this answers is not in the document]
gidsg: I removed the dependency injection and forced pushed (as it didn't seem to be adding any value). Would you mind re-reviewing @sfount?

sfount (approved):

PR #24, opened 2021-11-02
PP-8814 Database migration to add webhooks_messages table

alexbishop1:
A webhook message is a representation of what we will actually send to a service that has a webhook configured with a subscription to a particular event, combining data from the event itself and the transaction to which the event relates.

### DIFF
--- a/src/main/resources/migrations/0004_create_table_webhook_messages.sql
+++ b/src/main/resources/migrations/0004_create_table_webhook_messages.sql
@@ -1,0 +1,17 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:create-table-webhook_messages
+
+CREATE table webhook_messages (
+    id BIGSERIAL PRIMARY KEY,
+    external_id VARCHAR(30) NOT NULL,
+    created_date TIMESTAMP WITH TIME ZONE NOT NULL,
+    webhook_id INT NOT NULL,
+    event_date TIMESTAMP WITH TIME ZONE NOT NULL,
+    event_type INT NOT NULL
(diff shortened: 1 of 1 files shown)


src/main/resources/migrations/0004_create_table_webhook_messages.sql
```diff
@@ -0,0 +1,17 @@
...
+    id BIGSERIAL PRIMARY KEY,
+    external_id VARCHAR(30) NOT NULL,
+    created_date TIMESTAMP WITH TIME ZONE NOT NULL,
+    webhook_id INT NOT NULL,
+    event_date TIMESTAMP WITH TIME ZONE NOT NULL,
+    event_type INT NOT NULL
+);
+
+ALTER TABLE webhook_messages ADD CONSTRAINT fk_webhook_id FOREIGN KEY (webhook_id) REFERENCES webhooks (id);
+ALTER TABLE webhook_messages ADD CONSTRAINT fk_webhook_type_id FOREIGN KEY (event_type) REFERENCES event_types (id);
```
sfount, line 15: Very minor but to not conflate this with the similarly named `webhooks` table, similar could be done to the above name but I think that's much clearer 
```suggestion
ALTER TABLE webhook_messages ADD CONSTRAINT fk_webhook_message_event_type_id FOREIGN KEY (event_type) REFERENCES event_types (id);
```

alexbishop1: Sure

sfount (approved):

PR #25, opened 2021-11-11
PP-8814 Update dependencies in POM

alexbishop1:
- Explicitly depend on `jackson-core`, `jackson-annotations` and `jackson-databind` because Dropwizard depends on older versions of these so if we do not explicitly depend on all of them we end up with a confusing version mix
- Use bills of materials (BOMs) to avoid having to mention version numbers every time
- Update everything to the latest and greatest
- Reorder dependencies to be more logical


mrlumbu (approved):
Looks much nicer. And tests pass. So 👍

PR #26, opened 2021-11-12
PP-8814 Add Ledger client

mrlumbu:
Ledger client to return transaction by id.
Uses new Http client from standard lib.

### DIFF
--- a/src/main/java/uk/gov/pay/webhooks/ledger/LedgerService.java
+++ b/src/main/java/uk/gov/pay/webhooks/ledger/LedgerService.java
@@ -1,0 +1,46 @@
+package uk.gov.pay.webhooks.ledger;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import uk.gov.pay.webhooks.app.WebhooksConfig;
+import uk.gov.pay.webhooks.ledger.model.LedgerTransaction;
+
+import javax.inject.Inject;
+import java.io.IOException;
+import java.net.URI;
+import java.net.http.HttpClient;
+import java.net.http.HttpRequest;
+import java.net.http.HttpResponse;
+import java.util.Optional;
+
+import static java.nio.charset.StandardCharsets.UTF_8;
+
+public class LedgerService {
+
+    private final HttpClient httpClient;
+    private final String ledgerUrl;
+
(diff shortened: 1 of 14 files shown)


alexbishop1 (dismissed):
I paired with Dom on this and am confident it needs no further review under our software development policy.

alexbishop1 (approved):

src/main/java/uk/gov/pay/webhooks/ledger/LedgerService.java
```diff
@@ -0,0 +1,46 @@
...
+
+    public Optional<LedgerTransaction> getTransaction(String id) throws IOException, InterruptedException {
+        var uri = URI.create(ledgerUrl + "/v1/transaction/" + id + "?override_account_id_restriction=true");
+        return getTransactionFromLedger(uri);
+    }
+
+    private Optional<LedgerTransaction> getTransactionFromLedger(URI uri) throws IOException, InterruptedException {
+        HttpResponse<String> response = getResponse(uri);
+        if (response.statusCode() == 200) {
+            var objectMapper = new ObjectMapper();
```
jamietanna, line 36: these can be quite costly to instantiate - can we either inject this, or make it a constant?

src/main/java/uk/gov/pay/webhooks/ledger/LedgerService.java
```diff
@@ -0,0 +1,46 @@
...
+            return Optional.of(objectMapper.readValue(response.body(), LedgerTransaction.class));
```
jamietanna, line 37: Is there any chance this could be nullable? :thinking:

src/main/java/uk/gov/pay/webhooks/ledger/LedgerService.java
mrlumbu: yeah, we will. And in fact have elsewhere. Thanks for picking this up

src/main/java/uk/gov/pay/webhooks/ledger/LedgerService.java
mrlumbu: maybe. Worth allowing for 👍

PR #27, opened 2021-11-16
PP-8814 Filter internal events and build webhook message

mrlumbu:
Adds webhook message service with method `handleInternalEvent`
that will be used to process incoming internal events.
At the moment the method finds webhooks subscribed to
the event, and then for each such webhook builds a webhook
message (including fetching transaction from ledger),
and stores it.

I have added a test for the filtering, but will leave off
writing an integration test until we have this working end
to end.

### DIFF
--- a/src/main/java/uk/gov/pay/webhooks/message/EventMapper.java
+++ b/src/main/java/uk/gov/pay/webhooks/message/EventMapper.java
@@ -1,0 +1,20 @@
+package uk.gov.pay.webhooks.message;
+
+import uk.gov.pay.webhooks.eventtype.EventTypeName;
+
+import java.util.Map;
+import java.util.Optional;
+
+public class EventMapper {
+    private static final Map<String, EventTypeName> internalToWebhook = Map.of(
+            "CAPTURE_CONFIRMED", EventTypeName.CARD_PAYMENT_CAPTURED
+    );
+
+    private static final Map<EventTypeName, String> webhookToInternal = Map.of(
+            EventTypeName.CARD_PAYMENT_CAPTURED, "CAPTURE_CONFIRMED"
+    );
+
+    public static Optional<String> getInternalEventNameFor(EventTypeName webhookEventTypeName) {
+        return Optional.ofNullable(webhookToInternal.get(webhookEventTypeName));
+    }
+}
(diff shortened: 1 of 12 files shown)


src/main/java/uk/gov/pay/webhooks/message/EventMapper.java
```diff
@@ -0,0 +1,20 @@
+package uk.gov.pay.webhooks.webhookevent;
```
gidsg, line 1: Package name doesn't match filename. Is this intentional or can you rename one or t'other?

src/main/java/uk/gov/pay/webhooks/webhook/WebhookService.java
```diff
@@ -79,5 +86,19 @@ public WebhookEntity update(String externalId, String serviceId, List<JsonPatchR
...
+        return list(event.live(), event.serviceId())
+                .stream()
+                .filter(webhook -> webhookHasSubscriptionForEvent(webhook, event))
+                .toList();
+    }
+
+    private boolean webhookHasSubscriptionForEvent(WebhookEntity webhook, InternalEvent event) {
+        return webhook.getSubscriptions().stream()
+                .map(EventTypeEntity::getName)
+                .map(uk.gov.pay.webhooks.webhookevent.EventMapper::getInternalEventNameFor)
```
gidsg, line 100: Can we use an `import` rather than including the fully-qualified name here?

src/main/java/uk/gov/pay/webhooks/message/dao/WebhookMessageDao.java
```diff
@@ -0,0 +1,19 @@
...
+import uk.gov.pay.webhooks.message.dao.entity.WebhookMessageEntity;
+
+import javax.inject.Inject;
+
+public class WebhookMessageDao extends AbstractDAO<WebhookMessageEntity> {
+
+    @Inject
+    public WebhookMessageDao(SessionFactory factory) {
+        super(factory);
+        }
```
gidsg, line 14: Can we add a line break between the methods here?

src/main/java/uk/gov/pay/webhooks/message/EventMapper.java
mrlumbu: 👍

src/main/java/uk/gov/pay/webhooks/webhook/WebhookService.java
mrlumbu: 👍

src/main/java/uk/gov/pay/webhooks/message/dao/WebhookMessageDao.java
mrlumbu: 👍

gidsg (approved):
👍 🚀

PR #29, opened 2021-12-09
PP-8813 Send webhook message with signature header

alexbishop1:
Not-yet-hooked-up-to-anything code to send a webhook message to the configured callback URL. The request includes an HTTP header of the following form:

```
Pay-Signature: <SHA-256 HMAC of body using configured secret>
```

The expected signature in `WebhookMessageSignatureGenerator` has been verified using two online HMAC generators.

`WebhookMessageSender` just concerns itself with sending the webhook. The response (which may have an HTTP status code indicating an error) is returned to the caller and exceptions are allowed to propagate. The future calling code will do something sensible to handle any errors.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -48,7 +48,7 @@
             <dependency>
                 <groupId>org.junit</groupId>
                 <artifactId>junit-bom</artifactId>
-                <version>5.8.1</version>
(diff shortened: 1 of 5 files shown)


gidsg (approved):
LGTM 🖊️ 👍

PR #30, opened 2021-12-14
PP-9001: webhook delivery queue entity

gidsg:
This splits out the new entity for the webhook delivery queue, so we can start building views for it, separate from the sending functionality.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -98,6 +98,11 @@
 
         <!-- Main dependencies that need explicit versions -->
         <dependency>
+            <groupId>com.vladmihalcea</groupId>
+            <artifactId>hibernate-types-55</artifactId>
+            <version>2.14.0</version>
+        </dependency>
+        <dependency>
             <groupId>org.liquibase</groupId>
             <artifactId>liquibase-core</artifactId>
             <version>4.4.3</version>
--- a/src/main/java/uk/gov/pay/webhooks/deliveryqueue/dao/WebhookDeliveryQueueEntity.java
+++ b/src/main/java/uk/gov/pay/webhooks/deliveryqueue/dao/WebhookDeliveryQueueEntity.java
@@ -1,0 +1,139 @@
+package uk.gov.pay.webhooks.deliveryqueue.dao;
+
+import uk.gov.pay.webhooks.message.dao.entity.WebhookMessageEntity;
+
+import javax.persistence.Column;
+import javax.persistence.Entity;
+import javax.persistence.GeneratedValue;
+import javax.persistence.GenerationType;
+import javax.persistence.Id;
+import javax.persistence.JoinColumn;
+import javax.persistence.ManyToOne;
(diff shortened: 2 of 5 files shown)


src/main/java/uk/gov/pay/webhooks/deliveryqueue/dao/WebhookDeliveryQueueEntity.java
```diff
@@ -0,0 +1,139 @@
...
+        FAILED
+    }
+
+
+    public void setDeliveryStatus(String deliveryStatus) {
+        this.deliveryStatus = deliveryStatus;
+    }
+
+
+    public String getDeliveryStatus() {
```
mrlumbu, line 73: would be nice to only expose enum here

src/main/java/uk/gov/pay/webhooks/deliveryqueue/dao/WebhookDeliveryQueueEntity.java
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    @Column(name = "delivery_result")
+    private String deliveryResult;
+
+    public Integer getStatusCode() {
+        return statusCode;
+    }
+
+    public void setStatusCode(Integer statusCode) {
```
mrlumbu, line 95: guess this will be clearer when you push context for this, but what is this status code? Could you explain briefly in PR (and commit)

pom.xml
```diff
@@ -97,6 +97,11 @@
         </dependency>
 
         <!-- Main dependencies that need explicit versions -->
+        <dependency>
+            <groupId>com.vladmihalcea</groupId>
+            <artifactId>hibernate-types-55</artifactId>
+            <version>2.14.0</version>
+        </dependency>
```
mrlumbu, line 104: could you explain the purpose of introducing this dependency in PR

mrlumbu (changes requested):

src/main/java/uk/gov/pay/webhooks/deliveryqueue/dao/WebhookDeliveryQueueEntity.java
gidsg: @mrlumbu This is for the HTTP response code returned by the attempt (if any). Do you think we could have a clearer name to make that clearer?

src/main/java/uk/gov/pay/webhooks/deliveryqueue/dao/WebhookDeliveryQueueEntity.java
gidsg: 👍  have done this now

pom.xml
gidsg: @mrlumbu I have attempted to explain this in the extended commit message.

mrlumbu (approved):

PR #31, opened 2021-12-14
PP-8986 Read internal events from queue

alexbishop1:
Read internal events (from elsewhere in the platform) from the SQS queue and process them to send webhook messages.

### DIFF
--- a/src/main/java/uk/gov/pay/webhooks/app/SqsConfig.java
+++ b/src/main/java/uk/gov/pay/webhooks/app/SqsConfig.java
@@ -1,0 +1,58 @@
+package uk.gov.pay.webhooks.app;
+
+import io.dropwizard.Configuration;
+
+import javax.validation.constraints.Max;
+import javax.validation.constraints.NotNull;
+
+public class SqsConfig extends Configuration {
+
+    @NotNull
+    private String eventQueueUrl;
+
+    @NotNull
+    private String region;
+    private String accessKey;
+    private String secretKey;
+
+    @Max(20)
+    private int messageMaximumWaitTimeInSeconds;
+    @Max(10)
+    private int messageMaximumBatchSize;
+
+
+    private boolean nonStandardServiceEndpoint;
+    private String endpoint;
+
+    public String getEventQueueUrl() {
+        return eventQueueUrl;
+    }
+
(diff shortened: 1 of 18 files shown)


src/main/java/uk/gov/pay/webhooks/queue/QueueMessageReceiver.java
```diff
@@ -0,0 +1,70 @@
...
+                .threads(queueReadScheduleNumberOfThreads)
+                .build();
+    }
+
+    @Override
+    public void start() {
+        long initialDelay = config.getThreadDelayInMilliseconds();
+        long delay = config.getThreadDelayInMilliseconds();
+
+        for(int i = 0; i < queueReadScheduleNumberOfThreads; i++) {
```
gidsg, line 46: could we do this as a forEach instead?

gidsg (approved):
Looks good to me, aside from minor comment 👍 👑

src/main/java/uk/gov/pay/webhooks/app/SqsConfig.java
```diff
@@ -0,0 +1,58 @@
...
+    private String accessKey;
+    private String secretKey;
+
+    @Max(20)
+    private int messageMaximumWaitTimeInSeconds;
+    @Max(10)
+    private int messageMaximumBatchSize;
+
+
+    private boolean nonStandardServiceEndpoint;
```
sfount, line 24: I think using language like "AWS environment" or "local environment" might help future us understand this without as much README. No suggestions at this hour, happy for it to be in a follow up after discussion

src/main/java/uk/gov/pay/webhooks/queue/QueueMessageReceiver.java
alexbishop1: Not really. `queueReadScheduleNumberOfThreads` is just an `int`. There’s no collection we’re iterating over here.

src/main/java/uk/gov/pay/webhooks/app/SqsConfig.java
alexbishop1: Yeah, let’s change this in all the queue-consuming projects at once.

PR #32, opened 2021-12-14
PP-9001: add sending task

gidsg:
DRAFT


sfount (approved):
Great first pass 👍 we've addressed a couple of issues in this change set locally which will follow as a PR immediately following

PR #37, opened 2022-01-06
PP-9001 Patch Webhook send process

sfount:
See commit messages for details. Address a number of small issues with the polling/ emitting webhook messages processes.

- add database transaction session for polling
- fix emission database field date granularity
- inject sqs client for guice modules
- parse additional SNS/ SQS message layers

### DIFF
--- a/src/main/java/uk/gov/pay/webhooks/deliveryqueue/managed/WebhookMessageSendingQueueProcessor.java
+++ b/src/main/java/uk/gov/pay/webhooks/deliveryqueue/managed/WebhookMessageSendingQueueProcessor.java
@@ -66,31 +66,32 @@ public class WebhookMessageSendingQueueProcessor implements Managed {
 
     public void processQueue() {
         try {
-            Session session = sessionFactory.openSession();
-            pollQueue(session);
-
+            pollQueue();
         } catch (Exception e) {
+            LOGGER.warn("Failed to poll queue %s".formatted(e.getMessage()));
             e.printStackTrace();
         }
-
-
     }
 
     private void attemptSend(WebhookDeliveryQueueEntity queueItem) {
         var retryCount = webhookDeliveryQueueDao.countFailed(queueItem.getWebhookMessageEntity());
-        
+
         var webhookMessageSender = new WebhookMessageSender(httpClient, objectMapper, webhookMessageSignatureGenerator);
         try {
             LOGGER.info("Attempting to send Webhook ID %s to %s".formatted(queueItem.getWebhookMessageEntity().getExternalId(), queueItem.getWebhookMessageEntity().getWebhookEntity().getCallbackUrl()));
             var response = webhookMessageSender.sendWebhookMessage(queueItem.getWebhookMessageEntity());
+
             var statusCode = response.statusCode();
             if (statusCode >= 200 && statusCode <= 299) {
+                LOGGER.info("Message attempt succeeded with %s".formatted(statusCode));
                 webhookDeliveryQueueDao.recordResult(queueItem, String.valueOf(statusCode), statusCode, WebhookDeliveryQueueEntity.DeliveryStatus.SUCCESSFUL);
             } else {
+                LOGGER.info("Message attempt failed with %s".formatted(statusCode));
                 webhookDeliveryQueueDao.recordResult(queueItem, String.valueOf(statusCode), statusCode, WebhookDeliveryQueueEntity.DeliveryStatus.FAILED);
                 enqueueRetry(queueItem, nextRetryIn(retryCount));
             }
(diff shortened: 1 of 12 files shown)


src/test/java/uk/gov/pay/webhooks/queue/EventQueueIT.java
```diff
@@ -50,18 +50,19 @@ public void shouldReceiveMessageFromTheQueue() throws QueueException {
...
-                	"timestamp": "2018-03-12T16:25:01.123456Z",
-                	"resource_external_id": "3uwuyr38rry",
-                	"eventType": "PAYMENT_CREATED",
-                	"event_details": {
-                		"example_event_details_field": "and its value"
-                	}
-                }""";
+                  "Type" : "Notification",
+                  "MessageId" : "04424f78-d540-5eb7-87e1-154d586b6b02",
+                  "Message" : "{\\\"sqs_message_id\\\":\\\"dc142884-1e4b-4e57-be93-111b692a4868\\\",\\\"service_id\\\":\\\"some-service-id\\\",\\\"live\\\":false,\\\"resource_type\\\":\\\"payment\\\",\\\"resource_external_id\\\":\\\"t8cj9v1lci7da7pbp99qg9olv3\\\",\\\"parent_resource_external_id\\\":null,\\\"event_date\\\":\\\"2019-08-31T14:18:46.446541Z\\\",\\\"event_type\\\":\\\"PAYMENT_DETAILS_ENTERED\\\",\\\"event_data\\\":{},\\\"reproject_domain_object\\\":false}",
```
gidsg, line 57:
```suggestion
                  "Message" : "{\\"sqs_message_id\\":\\"dc142884-1e4b-4e57-be93-111b692a4868\\",\\"service_id\\":\\"some-service-id\\",\\"live\\":false,\\"resource_type\\":\\"payment\\",\\"resource_external_id\\":\\"t8cj9v1lci7da7pbp99qg9olv3\\",\\"parent_resource_external_id\\":null,\\"event_date\\":\\"2019-08-31T14:18:46.446541Z\\",\\"event_type\\":\\"PAYMENT_DETAILS_ENTERED\\",\\"event_data\\":{},\\"reproject_domain_object\\":false}",
```
IntelliJ says some of this escaping is redundant, so this should be equivalent.

src/main/java/uk/gov/pay/webhooks/deliveryqueue/managed/WebhookMessageSendingQueueProcessor.java
```diff
@@ -66,38 +66,44 @@ public void start() {
...
-            LOGGER.warn("Unexpected exception %s attempting to send webhook message ID: %s".formatted(e.getMessage(), queueItem.getWebhookMessageEntity().getExternalId()));
+            LOGGER.warn("Exception %s attempting to send webhook message ID: %s".formatted(e.getMessage(), queueItem.getWebhookMessageEntity().getExternalId()));
             webhookDeliveryQueueDao.recordResult(queueItem, e.getMessage(), null, WebhookDeliveryQueueEntity.DeliveryStatus.FAILED);
             enqueueRetry(queueItem, nextRetryIn(retryCount));
-            
+        } catch (Exception e) {
+            // handle all exceptions at this level to make sure that the retry mechanism is allowed to work as designed
+            // allowing errors passed this point (not guaranteeing an update) would allow perpetual failures 
+            LOGGER.warn("Unexpected exception %s attempting to send webhook message ID: %s".formatted(e.getMessage(), queueItem.getWebhookMessageEntity().getExternalId()));
+            webhookDeliveryQueueDao.recordResult(queueItem, "Unknown error", null, WebhookDeliveryQueueEntity.DeliveryStatus.FAILED);
```
gidsg, line 105: Maybe it would be better to propagate the exception message up here? On the other hand I guess no guarantee it would be ledgeble to an external user.

gidsg (approved):
LGTM, thanks for figuring out these fixes 🌟 👍 
(Would be nice to simplify the serialisation steps a bit, but this could be done potentially in a separate PR)

src/main/java/uk/gov/pay/webhooks/deliveryqueue/managed/WebhookMessageSendingQueueProcessor.java
sfount: Hmm yeah I had considered this but am a bit concerned about exposing system level messages out to consumers without some kind of map. I can't think how but it feels like there could be some sharp edges or unintended behaviour. Not totally against it, it will be good to refine these delivery status. messages once the pieces are in place.

sfount:
@gidsg agreed a nested serialisation would clean things up significantly! happy to push a separate PR or for you to include in your next changes

PR #38, opened 2022-01-06
BAU Allow webhooks logs to go through to Splunk

sfount:
The lambda responsible for ingesting output from fargate instances
requires an `environment` attribute to be set (using commons structured
logging). This sets the `environment` and `container` attributes for all
messages logged by the service which will allow these lines to be
ingested.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -74,6 +74,10 @@
         </dependency>
         <dependency>
             <groupId>io.dropwizard</groupId>
+            <artifactId>dropwizard-json-logging</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.dropwizard</groupId>
             <artifactId>dropwizard-migrations</artifactId>
             <exclusions>
                 <exclusion>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -72,6 +72,15 @@
...
         </dependency>
+        <dependency>
+            <groupId>io.dropwizard</groupId>
+            <artifactId>dropwizard-json-logging</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>uk.gov.service.payments</groupId>
+            <artifactId>logging</artifactId>
+            <version>${pay-java-commons.version}</version>
+        </dependency>
```
alexbishop1, line 83: Can you move this into the ‘Main dependencies that need explicit versions’ section please?

alexbishop1 (changes requested):
Looks good except for one tiny thing.

sfount:
@alexbishop1 thanks for the review, moved the commons dependency and kept the dropwizard where it is (no explicit versions needed)

PR #41, opened 2022-01-07
PP-9001: Extract attempt sending logic in separate SendAttempter class

gidsg:
Extracts the sending & retry logic from the scheduled task execution to make it more test-able.

* Extract attempt sending logic in separate SendAttempter class
* Prettify delivery result message with HTTP reason
* Inject WebhookMessageSender into SendAttempter to allow for mocking in tests
* Add tests for SendAttempter

### DIFF
--- a/src/main/java/uk/gov/pay/webhooks/deliveryqueue/managed/SendAttempter.java
+++ b/src/main/java/uk/gov/pay/webhooks/deliveryqueue/managed/SendAttempter.java
@@ -1,0 +1,92 @@
+package uk.gov.pay.webhooks.deliveryqueue.managed;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import uk.gov.pay.webhooks.deliveryqueue.dao.WebhookDeliveryQueueDao;
+import uk.gov.pay.webhooks.deliveryqueue.dao.WebhookDeliveryQueueEntity;
+import uk.gov.pay.webhooks.message.WebhookMessageSender;
+
+import javax.inject.Inject;
+import javax.ws.rs.core.Response;
+import java.io.IOException;
+import java.net.http.HttpTimeoutException;
+import java.security.InvalidKeyException;
+import java.time.Duration;
+import java.time.InstantSource;
+import java.time.temporal.ChronoUnit;
+import java.util.Date;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+public class SendAttempter {
+    private static final Logger LOGGER = LoggerFactory.getLogger(SendAttempter.class);
+    private WebhookDeliveryQueueDao webhookDeliveryQueueDao;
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/pay/webhooks/deliveryqueue/managed/WebhookMessageSendingQueueProcessor.java
```diff
@@ -135,7 +85,7 @@ private void pollQueue() {
         Transaction transaction = session.beginTransaction();
         try {
             Optional<WebhookDeliveryQueueEntity> maybeQueueItem = webhookDeliveryQueueDao.nextToSend(Date.from(instantSource.instant()));
-            maybeQueueItem.ifPresent(this::attemptSend);
+            maybeQueueItem.ifPresent(item -> new SendAttempter(webhookDeliveryQueueDao, instantSource, new WebhookMessageSender(httpClient, objectMapper, webhookMessageSignatureGenerator)).attemptSend(item));
```
sfount, line 88: If I understand correctly, this is instantiating two fairly involved objects, per webhook message delivery attempt. I don't think this will be an issue for low volumes but given peak traffic I could see the time/ complexity adding up.

As this class is importing a lot of things and then just passing them down into instantiations it feels like these should use the same singleton pattern and have those imports injected further down rather than being passed down through instantiation.

Practically I think the `SendAttempter` can be injected as a singleton into this object, it can then be responsible for importing the `ObjectMapper`, `WebhookMessageSignatureGenerator` and `HTTPClient` (none of which are actually used in this object). 

I think the same kind of analysis can be done on the `httpClient` which is then initialised by the send attempter. I assume it's fine to be using the same http client for all of these requests but that's something that would need validating.

gidsg: @sfount Good points, hopefully have addressed in 738b3a0

src/main/java/uk/gov/pay/webhooks/deliveryqueue/managed/SendAttempter.java
```diff
@@ -0,0 +1,92 @@
...
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+public class SendAttempter {
+    private static final Logger LOGGER = LoggerFactory.getLogger(SendAttempter.class);
+    private WebhookDeliveryQueueDao webhookDeliveryQueueDao;
+    private InstantSource instantSource;
+    private WebhookMessageSender webhookMessageSender;
+
+    @Inject
```
sfount, line 30: I think this annotation can be removed if it will be initialised manually as in https://github.com/alphagov/pay-webhooks/pull/41/files#diff-bb525dc22ee461ff2cdcee4f8b1668836ea1dc55b47b754bf81cf36aaf76ae4fR43. 

I suspect it would be okay to let the DI manage this but I don't have an exact proposal here, happy to review after all the pieces are in

sfount (approved):
Much clearer abstraction 👍

PR #43, opened 2022-01-11
PP-8557 Webhook messages endpoint

sfount:
Adds endpoints for interrogating webhook messages and their delivery attempts.

---

`/v1/webhook/{externalId}/message`
List the messages for a given webhook
- ordered by created date
- 10 per page

`?page` - optional, specify how far into the data to fetch messages
`?status` - optional, filter by a given status `"SUCCESSFUL"`, `"FAILED"`

---

`/v1/webhook/{externalId}/message/{messageId}`
Get details for a given webhook and message

---

`/v1/webhook/{externalId}/message/{messageId}/attempt`
List all delivery attempts for a given webhook and message

### DIFF
--- a/src/main/java/uk/gov/pay/webhooks/message/dao/WebhookMessageDao.java
+++ b/src/main/java/uk/gov/pay/webhooks/message/dao/WebhookMessageDao.java
@@ -21,6 +21,13 @@ public class WebhookMessageDao extends AbstractDAO<WebhookMessageEntity> {
         return webhookMessage;
     }
 
+    public WebhookMessageEntity get(String webhookId, String messageId) {
+       return namedTypedQuery(WebhookMessageEntity.MESSAGE_BY_WEBHOOK_ID_AND_MESSAGE_ID)
+               .setParameter("webhookId", webhookId)
+               .setParameter("messageId", messageId)
+               .getSingleResult();
+    }
+
     public List<WebhookMessageEntity> list(String webhookId, String deliveryStatus, int page) {
         var query = (deliveryStatus != null) ?
                 namedTypedQuery(WebhookMessageEntity.MESSAGES_BY_WEBHOOK_ID_AND_STATUS).setParameter("webhookId", webhookId).setParameter("deliveryStatuses", List.of(deliveryStatus)) :
--- a/src/main/java/uk/gov/pay/webhooks/message/dao/WebhookMessageDao.java
+++ b/src/main/java/uk/gov/pay/webhooks/message/dao/WebhookMessageDao.java
@@ -3,33 +3,41 @@ package uk.gov.pay.webhooks.message.dao;
 import io.dropwizard.hibernate.AbstractDAO;
 import org.hibernate.SessionFactory;
 import uk.gov.pay.webhooks.message.dao.entity.WebhookMessageEntity;
-import uk.gov.pay.webhooks.webhook.dao.entity.WebhookEntity;
 
 import javax.inject.Inject;
 import java.util.List;
 
 public class WebhookMessageDao extends AbstractDAO<WebhookMessageEntity> {
 
+    public static final int WEBHOOK_MESSAGES_PAGE_SIZE = 10;
+
     @Inject
     public WebhookMessageDao(SessionFactory factory) {
         super(factory);
         }
-        
+
     public WebhookMessageEntity create(WebhookMessageEntity webhookMessage) {
         persist(webhookMessage);
         return webhookMessage;
     }
 
-    public List<WebhookMessageEntity> list(String webhookId) {
-        return namedTypedQuery(WebhookMessageEntity.MESSAGES_BY_WEBHOOK_ID)
-                .setParameter("webhookId", webhookId)
+    public List<WebhookMessageEntity> list(String webhookId, String deliveryStatus, int page) {
+        var query = (deliveryStatus != null) ?
+                namedTypedQuery(WebhookMessageEntity.MESSAGES_BY_WEBHOOK_ID_AND_STATUS).setParameter("webhookId", webhookId).setParameter("deliveryStatuses", List.of(deliveryStatus)) :
+                namedTypedQuery(WebhookMessageEntity.MESSAGES_BY_WEBHOOK_ID).setParameter("webhookId", webhookId);
+        return query.setFirstResult(calculateFirstResult(page))
+                .setMaxResults(WEBHOOK_MESSAGES_PAGE_SIZE)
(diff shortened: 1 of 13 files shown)


src/main/java/uk/gov/pay/webhooks/message/dao/WebhookMessageDao.java
```diff
@@ -5,16 +5,46 @@
...
+    public WebhookMessageEntity get(String webhookId, String messageId) {
+       return namedTypedQuery(WebhookMessageEntity.MESSAGE_BY_WEBHOOK_ID_AND_MESSAGE_ID)
+               .setParameter("webhookId", webhookId)
+               .setParameter("messageId", messageId)
+               .getSingleResult();
+    }
+
+    public List<WebhookMessageEntity> list(String webhookId, String deliveryStatus, int page) {
+        var query = (deliveryStatus != null) ?
+                namedTypedQuery(WebhookMessageEntity.MESSAGES_BY_WEBHOOK_ID_AND_STATUS).setParameter("webhookId", webhookId).setParameter("deliveryStatuses", List.of(deliveryStatus)) :
```
gidsg, line 33: Does this need to be a list if the endpoint only supports querying for one status?

src/test/java/uk/gov/pay/webhooks/message/dao/WebhookMessageDaoTest.java
```diff
@@ -0,0 +1,102 @@
...
+    public void setUp() {
+        webhookDao = new WebhookDao(database.getSessionFactory());
+        webhookMessageDao = new WebhookMessageDao(database.getSessionFactory());
+        webhookDeliveryQueueDao = new WebhookDeliveryQueueDao(database.getSessionFactory(), new InstantSource() {
+            @Override
+            public Instant instant() {
+                return Instant.now();
+            }
+        });
+    }
```
gidsg, line 46: we can simplify the creation of the InstantSource:
```suggestion
        webhookDeliveryQueueDao = new WebhookDeliveryQueueDao(database.getSessionFactory(), InstantSource.fixed(Instant.now()));
```

src/main/java/uk/gov/pay/webhooks/message/dao/WebhookMessageDao.java
sfount: hm no, this is an artefact of my initial methodology which was to have one query and either list all statuses (for no filter) or only one status for the filter. That ended up requiring more workarounds than it avoided because we intentionally don't include pending statuses in the `latest_attempt` field on webhook messages.

It happens to be the case the postgres query planner will convert `IS (1)` to `= 1` so there would be no appreciable difference so I'm happy to change this or leave it.

Will either make the change or not based on your vote!

src/test/java/uk/gov/pay/webhooks/message/dao/WebhookMessageDaoTest.java
sfount: Good catch, this is much nicer 👍

src/main/java/uk/gov/pay/webhooks/webhook/resource/WebhookResource.java
```diff
@@ -103,7 +107,39 @@ public SigningKeyResponse regenerateSigningKey(@PathParam("externalId") @NotNull
...
     }
-    
+
+    @UnitOfWork
+    @Path("/{externalId}/messages")
+    @GET
+    public WebhookMessageSearchResponse getWebhookMessages(
+            @PathParam("externalId") String externalId,
+            @QueryParam("page") Integer page,
+            @QueryParam("status") String status
```
gidsg, line 117: Can we fail the request if the `status` does not appear in the `DeliveryStatus` enum?

sfount: Yeah doing that explicitly sounds like a good idea. 

At the moment it will just return an empty set as nothing will have that status, i.e 
```
{ 
  count: 0,
  page: 1,
  total: 0,
  results: []
}
```

src/main/java/uk/gov/pay/webhooks/webhook/resource/WebhookResource.java
```diff
@@ -103,7 +107,39 @@ public SigningKeyResponse regenerateSigningKey(@PathParam("externalId") @NotNull
...
+    ) {
+        var currentPage = (page != null) ? page : 1;
```
gidsg, line 119:
```suggestion
        var currentPage = Optional.ofNullable(page).orElse(1);
```
maybe a bit nicer? But only a style thing!

src/main/java/uk/gov/pay/webhooks/message/dao/WebhookMessageDao.java
```diff
@@ -5,16 +5,46 @@
...
+
+    public WebhookMessageEntity get(String webhookId, String messageId) {
+       return namedTypedQuery(WebhookMessageEntity.MESSAGE_BY_WEBHOOK_ID_AND_MESSAGE_ID)
+               .setParameter("webhookId", webhookId)
+               .setParameter("messageId", messageId)
+               .getSingleResult();
+    }
+
+    public List<WebhookMessageEntity> list(String webhookId, String deliveryStatus, int page) {
+        var query = (deliveryStatus != null) ?
```
gidsg, line 32:
```suggestion
        var query = deliveryStatus != null ?
```
unnecessary parenthesis.

src/main/java/uk/gov/pay/webhooks/message/dao/WebhookMessageDao.java
gidsg: I think I prefer it without the list/`IN`, as it simplifies the code and the query (or at least I think made it easier in my head to parse what was going on). We can always add it back if we do add that functionality later.

src/main/java/uk/gov/pay/webhooks/webhook/resource/WebhookResource.java
sfount: Much prefer this, thanks 👍 including the above suggestion, I've gone with 

```java
var currentPage = Optional.ofNullable(page)
                .orElse(1);

 var deliveryStatus = Optional.ofNullable(status)
                .map(String::valueOf)
                .orElse(null);
```

I was out here trying to write functional JS 😅 
```javascript
const currentPage = page && 1
```

gidsg (approved):
📟  🌟 👍

PR #44, opened 2022-01-11
PP-9001: add webhook message model for serialised webhook message body

gidsg:
Addresses TODO to serialise response in expected format i.e. as described in: https://github.com/alphagov/pay-decisions/blob/20aeb20c6653ff8fbd27069f65ca85c11511e560/RFC/003-webhooks.md#appendix

Note: This just serialises the `LedgerTransaction` as-is within the `resource` field. Let me know if any further transformation is required

* add missing field 'resource_type' serialised from SQS message and added as top level field to message
* Serialises/Deserialises timestamp from SQS message as Instant


src/main/java/uk/gov/pay/webhooks/message/WebhookMessage.java
```diff
@@ -0,0 +1,42 @@
...
+import uk.gov.service.payments.commons.api.json.ApiResponseInstantSerializer;
+import uk.gov.service.payments.commons.api.json.MicrosecondPrecisionDateTimeDeserializer;
+
+import java.time.Instant;
+import java.time.ZonedDateTime;
+import java.util.Date;
+
+
+public record WebhookMessage(String id,
+                             @JsonProperty("created_date") @JsonSerialize(using = ApiResponseInstantSerializer.class) Instant createdDate,
```
sfount, line 23: I think unless this is being automatically created with an ObjectMapper (seem to recall there was some corner case there) all of these explicit property annotations can be replaced by annotating the record with 
```java
@JsonNaming(PropertyNamingStrategies.SnakeCaseStrategy.class)
```

src/main/java/uk/gov/pay/webhooks/message/WebhookMessageService.java
```diff
@@ -68,15 +68,15 @@ public void handleInternalEvent(InternalEvent event) throws IOException, Interru
...
 
         var webhookMessageEntity = new WebhookMessageEntity();
         webhookMessageEntity.setExternalId(idGenerator.newExternalId());
         webhookMessageEntity.setCreatedDate(Date.from(instantSource.instant()));
         webhookMessageEntity.setWebhookEntity(webhook);
-        webhookMessageEntity.setEventDate(Date.from(event.eventDate().toInstant()));
+        webhookMessageEntity.setEventDate(Date.from(event.eventDate()));
         webhookMessageEntity.setEventType(eventTypeDao.findByName(EventMapper.getWebhookEventNameFor(event.eventType())).orElseThrow(IllegalArgumentException::new));
-        webhookMessageEntity.setResource(resource);
+        webhookMessageEntity.setResource(objectMapper.valueToTree(WebhookMessage.of(webhookMessageEntity, event, resource)));
```
sfount, line 79: Do we want to presist this formatted payload in the database or do we think it would be better to keep the resource stored separately and format this at the time of sending? I had initially thought the latter to preserve storing just the resource but happy to discuss pros/ cons 

If we did go with this method I think we should rename this property from the entity through to the database to make it clear it's a cache of the webhook payload not just the resource fetched from the event store

sfount (commented):
Looking good! One suggestion and a question, will review again after ddiscussion

src/main/java/uk/gov/pay/webhooks/message/WebhookMessage.java
gidsg: Good shout, have done this in 5ee32fa.

src/main/java/uk/gov/pay/webhooks/message/WebhookMessageService.java
gidsg: @sfount As discussed offline, we think it makes sense to persist the individual fields from the wrapper and then do the transform at send time. Maybe I can change this PR to just do the transform at send time, and then do a separate PR to persist the individual wrapper fields?

gidsg: @sfount ...OK I have done this now in this PR in de53ccf

src/main/java/uk/gov/pay/webhooks/util/MicrosecondPrecisionInstantDeserializer.java
```diff
@@ -0,0 +1,25 @@
...
+public class MicrosecondPrecisionInstantDeserializer extends JsonDeserializer<Instant> {
+
+    public static final DateTimeFormatter MICROSECOND_FORMATTER =
+            MicrosecondPrecisionInstantSerializer.MICROSECOND_FORMATTER;
+    
+
+    @Override
+    public Instant deserialize(JsonParser p, DeserializationContext ctxt)  {
+        try {
+            return Instant.from(MICROSECOND_FORMATTER.parse(p.getText()));
```
alexbishop1, line 19: I think we can get away without needing to rely on a `DateTimeFormatter` here: we can just use `Instant.parse(p.getText())`. This will will be slightly more liberal in terms of what it accepts but since everything it‘s parsing will be created by us, I think that’s OK. Without the `DateTimeFormatter` this class could then become `InstantDeserializer`, which will be more reusable.

src/main/java/uk/gov/pay/webhooks/util/MicrosecondPrecisionInstantDeserializer.java
```diff
@@ -0,0 +1,25 @@
...
+        } catch (Exception e) {
```
alexbishop1, line 20: Do we want to be so broad in terms of what we catch here?

src/main/java/uk/gov/pay/webhooks/util/MicrosecondPrecisionInstantSerializer.java
```diff
@@ -0,0 +1,28 @@
...
+import java.time.format.DateTimeFormatterBuilder;
+import java.util.Locale;
+
+public class MicrosecondPrecisionInstantSerializer extends JsonSerializer<Instant> {
+
+    public static final DateTimeFormatter MICROSECOND_FORMATTER =
+            new DateTimeFormatterBuilder()
+                    .appendInstant(6)
+                    .toFormatter(Locale.ENGLISH)
+                    .withZone(ZoneOffset.UTC);
```
alexbishop1, line 21: Don’t need `withZone(…)` here: the input has no time zone and the output is an ISO instant, which is always UTC.

src/main/resources/migrations/0006_add_fields_to_webhook_messages.sql
```diff
@@ -0,0 +1,9 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:add-field-to-webhook-messages
+
+ALTER TABLE webhook_messages
+ADD COLUMN resource_external_id VARCHAR(26),
+ADD COLUMN resource_type VARCHAR;
+
+--rollback DROP column resource_external_id, DROP column resource_type;
```
alexbishop1, line 9: This should be in a separate PR (we deploy database changes separately from code changes).

src/test/java/uk/gov/pay/webhooks/message/WebhookMessageBodyTest.java
```diff
@@ -0,0 +1,69 @@
...
+
+@ExtendWith(DropwizardExtensionsSupport.class)
+class WebhookMessageBodyTest {
+        
+    private ObjectMapper objectMapper;
+    private InstantSource instantSource;
+
+    @BeforeEach
+    public void setUp() {
+        instantSource = InstantSource.fixed(Instant.from(Date.from(Instant.parse("2019-10-01T08:25:24.00Z")).toInstant()));
```
alexbishop1, line 28:
```java
instantSource = InstantSource.fixed(Instant.parse("2019-10-01T08:25:24.00Z"));
```

src/test/java/uk/gov/pay/webhooks/message/WebhookMessageSenderTest.java
```diff
@@ -69,8 +81,14 @@ void setUp() throws JsonProcessingException, InvalidKeyException {
         webhookMessageEntity = new WebhookMessageEntity();
         webhookMessageEntity.setWebhookEntity(webhookEntity);
         webhookMessageEntity.setResource(jsonPayload);
+        webhookMessageEntity.setEventDate(Date.from(Instant.from(Instant.parse("2019-10-01T08:25:24.00Z"))));
```
alexbishop1, line 84:
```java
webhookMessageEntity.setEventDate(Date.from(Instant.parse("2019-10-01T08:25:24.00Z")));
```

alexbishop1 (approved):

alexbishop1 (changes requested):
Whoops, didn’t mean to press approve!

src/main/resources/migrations/0006_add_fields_to_webhook_messages.sql
gidsg: @alexbishop1 OK, I have changed it now to update the existing migration, as we are currently re-creating the DB during each deploy.

src/main/java/uk/gov/pay/webhooks/util/MicrosecondPrecisionInstantDeserializer.java
gidsg: got rid of this broad try catch now.

src/test/java/uk/gov/pay/webhooks/message/WebhookMessageSenderTest.java
alexbishop1: Still creating an `Instant` from an `Instant`…

alexbishop1 (changes requested):
Just one thing where we’re going round the houses with `Instant`s.

alexbishop1 (approved):

PR #55, opened 2022-01-17
PP-9077 Support all initially proposed payment event types

sfount:
Add entries for
* Payment started
* Payment succeeded
* Payment refunded

Clean up duplication in K/V maps.

Add an additional case to fetch a parent resource if the event relates to a
dependant resource.


gidsg (approved):
LGTM 🧒 👍

PR #57, opened 2022-01-18
PP-9113: serialise Public API representation of payment

gidsg:
Transforms ledger transaction to Public API representation of a payment (excluding the `_links` block)

Models are copy-pasted from Public API (and Connector for Payment State enum)

N.B. This leaves a TODO for Refunds which is assumed will require a separate transformation.


gidsg:
Closed in favour of #67

PR #63, opened 2022-01-20
PP-9107 Fix workflow permissions

mrlumbu:
Need `contents: write` to write the tag. Have restricted this to the one job that needs it, which is run after merge.


gidsg (approved):
✍️ 👍

PR #77, opened 2022-02-02
BAU — Upgrade Eclipse Temurin Java runtime from 17.0.1 to 17.0.2

alexbishop1:
Also switch from a JDK image to a JRE image, which should be smaller while still meeting our needs.


gidsg (approved):
🆙 👍

PR #79, opened 2022-02-02
PP-9113: Filter out unsupported resource types

gidsg:
If there's a resource type we don't currently have an API representation of, filter it out rather than returning the Ledger Transaction model

### DIFF
--- a/src/main/java/uk/gov/pay/webhooks/message/WebhookMessageService.java
+++ b/src/main/java/uk/gov/pay/webhooks/message/WebhookMessageService.java
@@ -22,6 +22,7 @@ import javax.inject.Inject;
 import java.io.IOException;
 import java.time.InstantSource;
 import java.util.Date;
+import java.util.Optional;
 
 public class WebhookMessageService {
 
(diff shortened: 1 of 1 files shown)


src/main/java/uk/gov/pay/webhooks/message/WebhookMessageService.java
```diff
@@ -64,17 +65,24 @@ public void handleInternalEvent(InternalEvent event) throws IOException, Interru
...
         }
     }
 
-    private WebhookMessageEntity buildWebhookMessage(WebhookEntity webhook, InternalEvent event, LedgerTransaction ledgerTransaction) {
+    private Optional<WebhookMessageEntity> buildWebhookMessage(WebhookEntity webhook, InternalEvent event, LedgerTransaction ledgerTransaction) {
         JsonNode resource = switch (event.resourceType()) {
             case "payment" -> objectMapper.valueToTree(PaymentApiRepresentation.of(ledgerTransaction));
             case "refund" -> objectMapper.valueToTree(RefundApiRepresentation.of(ledgerTransaction));
-            default -> objectMapper.valueToTree(ledgerTransaction);
+            default -> null;
```
alexbishop1, line 78: Hmm, feels a bit like we’re using null as a magic value. Since we actually want to change the flow control, maybe we should just use a switch statement, which could be done like this:

```java
private Optional<WebhookMessageEntity> buildWebhookMessage(WebhookEntity webhook, InternalEvent event, LedgerTransaction ledgerTransaction) {
    switch (event.resourceType()) {
        case "payment" -> {
            JsonNode resource = objectMapper.valueToTree(PaymentApiRepresentation.of(ledgerTransaction));
            return Optional.of(buildWebhookMessageEntity(webhook, event, resource));
        }
        case "refund" -> {
            JsonNode resource = objectMapper.valueToTree(RefundApiRepresentation.of(ledgerTransaction));
            return Optional.of(buildWebhookMessageEntity(webhook, event, resource));
        }
        default -> {
            LOGGER.info("Ignoring unsupported resource type %s".formatted(event.resourceType()));
            return Optional.empty();
        }
    }
}

private WebhookMessageEntity buildWebhookMessageEntity(WebhookEntity webhook, InternalEvent event, JsonNode resource) {
    var webhookMessageEntity = new WebhookMessageEntity();
    // …
    return webhookMessageEntity;
}
```

src/main/java/uk/gov/pay/webhooks/message/WebhookMessageService.java
```diff
@@ -64,17 +65,24 @@ public void handleInternalEvent(InternalEvent event) throws IOException, Interru
...
         };
+        
+        if (Optional.ofNullable(resource).isEmpty()) {
```
alexbishop1, line 81: May as well just do `resource == null` here!

alexbishop1 (changes requested):
The first comment is a radical suggestion of how to do this differently.

The second comment is a must-fix if you decide not to implement the first comment.

gidsg:
@alexbishop1 thanks for the suggestions. I've implemented your first suggestion now and force-pushed.
Please may you re-review?

src/main/java/uk/gov/pay/webhooks/message/WebhookMessageService.java
```diff
@@ -64,17 +65,30 @@ public void handleInternalEvent(InternalEvent event) throws IOException, Interru
...
 
-    private WebhookMessageEntity buildWebhookMessage(WebhookEntity webhook, InternalEvent event, LedgerTransaction ledgerTransaction) {
-        JsonNode resource = switch (event.resourceType()) {
-            case "payment" -> objectMapper.valueToTree(PaymentApiRepresentation.of(ledgerTransaction));
-            case "refund" -> objectMapper.valueToTree(RefundApiRepresentation.of(ledgerTransaction));
-            default -> objectMapper.valueToTree(ledgerTransaction);
+    private Optional<WebhookMessageEntity> buildWebhookMessage(WebhookEntity webhook, InternalEvent event, LedgerTransaction ledgerTransaction) {
+        return switch (event.resourceType()) {
+            case "payment" -> {
+                var resource = objectMapper.valueToTree(PaymentApiRepresentation.of(ledgerTransaction));
```
alexbishop1, line 77: I actually prefer `JsonNode` to `var` in this case because it’s not immediately obvious what `objectMapper.valueToTree(…)` returns.

src/main/java/uk/gov/pay/webhooks/message/WebhookMessageService.java
```diff
@@ -64,17 +65,30 @@ public void handleInternalEvent(InternalEvent event) throws IOException, Interru
...
+                yield Optional.of(buildWebhookMessageEntity(webhook, event, resource));
```
alexbishop1, line 78: First use of `yield` in the Pay codebase!

(It was almost `break`. Yes, you’d write `break Optional.of(buildWebhookMessageEntity(webhook, event, resource));`!)

alexbishop1 (approved):

src/main/java/uk/gov/pay/webhooks/message/WebhookMessageService.java
gidsg: Yes, makes sense, have amended and force-pushed now.

gidsg:
Have slightly amended and force pushed as-per your comment @alexbishop1. Please may you re-review?

alexbishop1 (approved):

PR #81, opened 2022-02-03
PP-9202 Improve how we model timestamps

alexbishop1:
In short, make the following so:

- Except in entity classes, all timestamps in the application are `Instant`s
- In entity classes (that is, classes that represent something in the database), all timestamps are `OffsetDateTime`s
- In the database, all columns holding timestamps are of the type `TIMESTAMP WITH TIME ZONE`

Longer explanation:

Up until now, the webhooks application used a mixture of `Instant`s, `Date`s and `ZonedDateTime`s to represent timestamps. In the database, timestamps were a mixture of `TIMESTAMP WITH TIME ZONE` and `TIMESTAMP WITHOUT TIME ZONE`.

`Instant` is clearly the best class to model timestamps in the application. While we used Instants in most places, in entity objects we used `Dat`e because Hibernate can natively persist this to the database. `Date` is essentially deprecated and only supports millisecond precision (as opposed to Instant, which supports nanosecond precision, though in practice the system clocks on most Linux systems will only proffer microsecond-precision timestamps). `ZonedDateTime` has historically been used in the Pay platform but almost always with a time zone of UTC (or the default time zone with the expectation it will be UTC), so `Instant` is a simpler alternative in almost all cases.

Ideally, we’d simply use `Instant` everywhere. However, JPA 2.2+ does not natively support mapping an Instant in entity classes to an appropriate database column type. Unlike EclipseLink, Hibernate does not seem to support using JPA attribute converters to transform to an appropriate type on the fly (or if it does, we cannot get it working). Therefore, we have to use a natively supported type in each entity class. There are two such types defined in JPA 2.2+ that make sense:

- `OffsetDateTime`, which maps to `TIMESTAMP WITH TIME ZONE`
- `LocalDateTime`, which maps to `TIMESTAMP WITHOUT TIME ZONE`

Looking from the database side, there is not really much difference between `TIMESTAMP WITH TIME ZONE` and `TIMESTAMP WITHOUT TIME ZONE`. Indeed, PostgreSQL stores them in the same way (a count of nanoseconds since the epoch) and the time zone for `TIMESTAMP WITH TIME ZONE` is always UTC (PostgreSQL does conversion from other time zones when reading and writing). The webhooks application and the Pay platform in general use a mixture of both: either `TIMESTAMP WITH TIME ZONE` (which is UTC) or `TIMESTAMP WITHOUT TIME ZONE` (which the application code assumes is UTC).

`LocalDateTime` has no time zone, which makes it less than ideal for storing timestamps. `OffsetDateTime` is more appropriate.

Therefore, in entity classes, we now use `OffsetDateTime` to store timestamps. In the matching database columns, we use `TIMESTAMP WITH TIME ZONE`.

Note that only the type of the field itself is `OffsetDateTime`. We do not allow this type to ‘escape’ the entity class. We eagerly convert to and from Instant when necessary (conversion between `Instant`s and UTC `OffsetDateTime`s is infallible and lossless).

As such, setters in entities take an `Instant` and covert to a UTC `OffsetDateTime`:

```java
public void setCreatedDate(Instant createdDate) {
    this.createdDate = OffsetDateTime.ofInstant(createdDate,
                                                ZoneOffset.UTC);
}
```

Getters always return an `Instant`:

```java
public Instant getCreatedDate() {
    return createdDate.toInstant();
}
```

There should be no observable behavioural changes as a result of this change, with one exception: when JSON from ledger is converted to a `SettlementSummary` object, the `capture_submit_time` property is now parsed as an `ISO_INSTANT` rather than an `ISO_ZONED_DATE_TIME`. Every `ISO_INSTANT` can be parsed as an `ISO_ZONED_DATE_TIME` but the reverse is not true. Previously, parsing a string like this would have been supported:

```
2021-04-01T13:00:00.000+01:00[Europe/London]
```

Now, only strings like this are supported:

```
2021-04-01T12:00:00.000Z
```

Ledger only produces strings like that the latter, so this change should have no effect.

### DIFF
--- a/src/main/java/uk/gov/pay/webhooks/ledger/model/SettlementSummary.java
+++ b/src/main/java/uk/gov/pay/webhooks/ledger/model/SettlementSummary.java
@@ -5,9 +5,9 @@ import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
-import uk.gov.service.payments.commons.api.json.ApiResponseDateTimeDeserializer;
+import uk.gov.pay.webhooks.util.InstantDeserializer;
 
-import java.time.ZonedDateTime;
+import java.time.Instant;
 
 import static uk.gov.service.payments.commons.model.ApiResponseDateTimeFormatter.ISO_INSTANT_MILLISECOND_PRECISION;
 
@@ -16,9 +16,9 @@ import static uk.gov.service.payments.commons.model.ApiResponseDateTimeFormatter
 @JsonIgnoreProperties(ignoreUnknown = true)
 public class SettlementSummary {
 
-    @JsonDeserialize(using = ApiResponseDateTimeDeserializer.class)
+    @JsonDeserialize(using = InstantDeserializer.class)
     @JsonProperty("capture_submit_time")
-    private ZonedDateTime captureSubmitTime;
+    private Instant captureSubmitTime;
     @JsonProperty("captured_date")
     private String capturedDate;
 
@@ -26,7 +26,7 @@ public class SettlementSummary {
         return (captureSubmitTime != null) ? ISO_INSTANT_MILLISECOND_PRECISION.format(captureSubmitTime) : null;
     }
 
-    public void setCaptureSubmitTime(ZonedDateTime captureSubmitTime) {
+    public void setCaptureSubmitTime(Instant captureSubmitTime) {
         this.captureSubmitTime = captureSubmitTime;
     }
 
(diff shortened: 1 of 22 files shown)


src/main/java/uk/gov/pay/webhooks/ledger/model/SettlementSummary.java
```diff
@@ -16,17 +16,17 @@
 @JsonIgnoreProperties(ignoreUnknown = true)
 public class SettlementSummary {
 
-    @JsonDeserialize(using = ApiResponseDateTimeDeserializer.class)
```
gidsg, line 19: What's the difference between this new deserialiser and the existing [InstantDeserializer](https://github.com/alphagov/pay-webhooks/blob/e8883fb449daaddb6563b1833438497422e8cdb0/src/main/java/uk/gov/pay/webhooks/util/InstantDeserializer.java)?

alexbishop1: The main difference is that I did not know `InstantDeserializer` exists! I’ll update the code to use that.

gidsg (approved):
LGTM 🕐  👍

PR #87, opened 2022-02-10
BAU: reflect that authorisation summary is optional in ledger payload

gidsg:
We are seeing an exception on Webhooks in Test:
```
uk.gov.pay.webhooks.ledger.model.AuthorisationSummary.getThreeDSecure()" because the return value of "uk.gov.pay.webhooks.ledger.model.LedgerTransaction.getAuthorisationSummary() is null" 
```
which suggests this field is not always present in the ledger payload. This PR makes it optional, so no NPE will be thrown if it is not present.


sfount (approved):
This building of things being parsed from JSON still feels somewhat like jacksons job but this change just makes accessing an existing thing nullable, looks good to me 👍

PR #95, opened 2022-02-15
PP-9275: Add sentry integration

gidsg:
Add sentry integration in a similar fashion to other Pay apps.


src/main/resources/config/config.yaml
```diff
@@ -19,6 +19,11 @@ logging:
       customFields:
         container: "webhooks"
         environment: ${ENVIRONMENT}
+    - type: sentry
+      threshold: ERROR
+      dsn: ${SENTRY_DSN:-https://example.com@dummy/1}
```
alexbishop1, line 24: 🤨

gidsg: @alexbishop1 I think this is maybe to stop it throwing if value is not set in the environment. 
I got it from https://github.com/alphagov/pay-ledger/blob/f449a1394fb65be5161206179d60ff19ffc33328/src/main/resources/config/config.yaml#L59 but open to any better ideas?

src/main/java/uk/gov/pay/webhooks/deliveryqueue/managed/WebhookMessageSendingQueueProcessor.java
```diff
@@ -76,6 +77,7 @@ private void pollQueue() {
             maybeQueueItem.ifPresent(sendAttempter::attemptSend);
             transaction.commit();
         } catch (Exception e) {
+            Sentry.captureException(e);
             LOGGER.warn("Unexpected exception when polling queue  %s: ".formatted(e.getMessage()));
```
sfount, line 81: If we consider this something that should be reported to Sentry would it be better to log it at an error level and have it propagate through with the dropwizard integration?

Same with the other example here, there may well be reasons to avoid doing that

gidsg: @sfount I agree that is a cleaner approach, have changed it to this suggestion and force-pushed.

sfount (approved):
LGTM 👍 lines up with other app configs

PR #97, opened 2022-02-16
BAU — Dependency configuration: use Mockito bill of materials

alexbishop1:
Mockito now publish a bill of materials so use that in our POM.

We currently only use one Mockito dependency so this does not make any real difference but it’s nice to have in case we add more in the future.


gidsg (approved):
LGTM 👍

PR #102, opened 2022-02-18
BAU Serialise resource id and type on webhook message

sfount:
The admin tool currently parses the `resource` attribute of the webhook
message to get the transaction id (to be linked to and referenced). This
shouldn't be relied on as the shape of the api may in the future be
configurable.

The WebhookMessageEntity already parses and stores the resource ID based
on internal pay messages. Expose this on the serialised record so that
it can be used by the admin tool consumer.


gidsg (approved):
LGTM (Maybe `resourceType` could be more like an Enum in future, but this matches the current model) 👍

PR #140, opened 2022-03-21
BAU Remove unused serialised event fields

sfount:
Remove reference (pseudo contracts) to data we don't need on events.
Content of events isn't used when processing webhooks so reflect that
with models.


stephencdaly (approved):
👍

gidsg (approved):
LGTM (I guess we could always add this back if there's a use case) 👍

PR #241, opened 2022-05-26
BAU Add basic indexes on webhook tables

sfount:
Use the queries defined in the entities to cover attributes used in gets
and searches with indexes.
- `WebhookEntity`
- `WebhookMessageEntity`
- `WebhookDeliveryQueueEntity`

Also include a unique index constraint on the subscription table to both
speed up the many to one join and to avoid code mistakenly subscribing a
webhook to any given event hook more than once.


alexbishop1 (approved):
Normally I’d request we use `CREATE INDEX CONCURRENTLY` but since this app is not live yet, it’s fine.

PR #251, opened 2022-05-31
PP-9201 Refactor message polling logic

sfount:
Extract webhook message polling logic from the Dropwizard `Managed` context to allow it to be individually tested outside of background processes and threads.

Use the `DropwizardExtensionsSupport` to allow unit testing the key scenarios required, this allows
* verbose scenario testing
* very quick tests
* a (roughly) identical environment to running the Dropwizard app

The proposed algorithm lines up with the original ticket 
* poll 

A future PR should add an integration test to assert across the polling and emit to wiremock.


src/main/java/uk/gov/pay/webhooks/deliveryqueue/managed/WebhookMessagePollingService.java
```diff
@@ -0,0 +1,55 @@
+package uk.gov.pay.webhooks.deliveryqueue.managed;
+
+import org.hibernate.SessionFactory;
+import org.hibernate.context.internal.ManagedSessionContext;
```
alexbishop1, line 4: The fact this has `internal` in the package name hierarchy concerns me. We do not wish to be relying on internal implementation details.

src/main/java/uk/gov/pay/webhooks/deliveryqueue/managed/WebhookMessagePollingService.java
```diff
@@ -0,0 +1,55 @@
...
+        this.sendAttempter = sendAttempter;
+        this.sessionFactory = sessionFactory;
+    }
+
+    public void pollWebhookMessageQueue() {
+        Optional<WebhookDeliveryQueueEntity> attemptCursor;
+
+        do {
+            attemptCursor = sendIfAvailable();
+        } while(attemptCursor.isPresent());
```
alexbishop1, line 30: I wonder if it might just be simpler for `sendIfAvailable()` to return a boolean (since we never seem to look at anything inside the `WebhookDeliveryQueueEntity`).

src/main/java/uk/gov/pay/webhooks/deliveryqueue/managed/WebhookMessagePollingService.java
```diff
@@ -0,0 +1,55 @@
...
+                            sendAttempter.attemptSend(webhookDeliveryQueueEntity);
+                            transaction.commit();
+                            return webhookDeliveryQueueEntity;
+                        });
+            } catch (Exception e) {
+                transaction.rollback();
+                return Optional.empty();
+            }
+        } finally {
+            session.close();
```
alexbishop1, line 51: I think `session` should work with try-with-resources.

src/main/java/uk/gov/pay/webhooks/queue/EventMessageHandler.java
```diff
@@ -44,14 +44,17 @@ private void processSingleMessage(EventMessage message) throws QueueException {
...
-            LOGGER.warn("Event message with ID %s scheduled for retry: %s".formatted(message.queueMessage().messageId(), e.getMessage()));
-            transaction.rollback();
-            eventQueue.scheduleMessageForRetry(message);
+            } catch (Exception e) {
+                LOGGER.warn("Event message with ID %s scheduled for retry: %s".formatted(message.queueMessage().messageId(), e.getMessage()));
+                transaction.rollback();
+                eventQueue.scheduleMessageForRetry(message);
+            }
         } finally {
+            session.close();
```
alexbishop1, line 57: Again, I think this might work with try-with-resources.

src/test/java/uk/gov/pay/webhooks/deliveryqueue/managed/WebhookMessagePollingServiceTest.java
```diff
@@ -0,0 +1,138 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.*;
+
+@ExtendWith(DropwizardExtensionsSupport.class)
+@ExtendWith(MockitoExtension.class)
+@MockitoSettings(strictness = Strictness.LENIENT)
+class WebhookMessagePollingServiceTest {
+
+    public DAOTestExtension database = DAOTestExtension.newBuilder()
```
alexbishop1, line 40: `public`?

src/test/java/uk/gov/pay/webhooks/deliveryqueue/managed/WebhookMessagePollingServiceTest.java
```diff
@@ -0,0 +1,138 @@
...
+    private WebhookMessageDao webhookMessageDao;
+    private WebhookDao webhookDao;
+    private WebhookMessagePollingService webhookMessagePollingService;
+    private WebhookMessageSender webhookMessageSenderMock;
+    private SendAttempter sendAttempter;
+    @Mock
+    private HttpResponse response;
+
+    @BeforeEach
+    public void setUp() throws IOException, InvalidKeyException, InterruptedException {
```
alexbishop1, line 58: Don’t need `public` with JUnit 5.

src/test/java/uk/gov/pay/webhooks/deliveryqueue/managed/WebhookMessagePollingServiceTest.java
```diff
@@ -0,0 +1,138 @@
...
+
+    @Test
+    public void shouldEmitMultipleValidQueueItemsInOnePoll() throws IOException, InvalidKeyException, InterruptedException {
+        setupOrderedDeliveryQueueWith(List.of("first-external-id", "second-external-id"));
+        webhookMessagePollingService.pollWebhookMessageQueue();
+
+        var captor = ArgumentCaptor.forClass(WebhookMessageEntity.class);
+        verify(webhookMessageSenderMock, times(2)).sendWebhookMessage(captor.capture());
+        assertEquals("first-external-id", captor.getAllValues().get(0).getExternalId());
+        assertEquals("second-external-id", captor.getAllValues().get(1).getExternalId());
```
alexbishop1, line 95: I suspect in-order verification with `argThat(…)` might work here (not tat there’s anything wrong with this).

src/test/java/uk/gov/pay/webhooks/deliveryqueue/managed/WebhookMessagePollingServiceTest.java
```diff
@@ -0,0 +1,138 @@
...
+        database.inTransaction(() -> {
+            webhookEntity.setCallbackUrl("http://a-callback-url.com");
+            webhookEntity.setServiceId("a-service-id");
+            webhookEntity.setLive(false);
+
+            webhookDao.create(webhookEntity);
+        });
+
+        for (int i = 0; i < messageIds.size(); i++) {
+            int finalI = i;
```
alexbishop1, line 122: It took me a while to realise `final` is an effectively final variable, rather than the last value `i` gets or something.

You _could_ do this:

```java
IntStream.range(0, messageIds.size()).forEach(i -> {
    var message = database.inTransaction(() -> {
        WebhookMessageEntity webhookMessageEntity = new WebhookMessageEntity();
        var id = messageIds.get(i);
        // …
        });
    // …
});
```

src/test/java/uk/gov/pay/webhooks/deliveryqueue/managed/WebhookMessagePollingServiceTest.java
```diff
@@ -0,0 +1,138 @@
...
+                WebhookMessageEntity webhookMessageEntity = new WebhookMessageEntity();
+                var id = messageIds.get(finalI);
+                webhookMessageEntity.setExternalId(id);
+                webhookMessageEntity.setWebhookEntity(webhookEntity);
+                webhookMessageEntity.setCreatedDate(instantSource.instant());
+
+                return webhookMessageDao.create(webhookMessageEntity);
+            });
+            database.inTransaction(() -> {
+                webhookDeliveryQueueDao.enqueueFrom(message, WebhookDeliveryQueueEntity.DeliveryStatus.PENDING, instantSource.instant().minusMillis((messageIds.size() - finalI) * 10L));
```
alexbishop1, line 134: I wonder if it might be clearer to have a mock `InputSource` and just stub it to return the desired value each time.

alexbishop1 (changes requested):
I like where the separation is going but I have concerns about using some Hibernate internals and there’s a couple of other suggestions.

alexbishop1 (changes requested):
Looks better. Just a couple of things.

src/test/java/uk/gov/pay/webhooks/deliveryqueue/managed/WebhookMessagePollingServiceTest.java
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+    @Test
+    public void shouldEmitSingleValidQueueItem() {
+        setupOrderedDeliveryQueueWith(List.of("first-external-id"));
+        database.inTransaction(() -> {
+            webhookMessagePollingService.pollWebhookMessageQueue();
+            try {
+                verify(webhookMessageSenderMock).sendWebhookMessage(argThat((webhookMessage -> webhookMessage.getExternalId().equals("first-external-id"))));
+            } catch (IOException | InvalidKeyException | InterruptedException ignored) { }
```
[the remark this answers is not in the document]
sfount: It's being executed in a lambda expression (call to db inTransaction) because of the nature of the unit test (hibernate doesn't other have an existing session as the apps not running or managing it), so unfortunately these errors don't belong to the test method signature -- it was like this before but now that we're no longer manually managing the hibernate transactions the lambda expression is needed here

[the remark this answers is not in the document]
alexbishop1: That makes sense. If we do not expect these exceptions to happen, stick a `fail()` call inside the catch block so that if they ever do happen, we immediately fail the test.

PR #254, opened 2022-06-01
PP-9201 Database migration to record webhook message response time

sfount:
Add an integer column to track response time from webhook delivery queue
message attempts.


src/main/resources/migrations/0009_alter_table_webhook_delivery_queue_add_column_duration_in_millis.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE webhook_delivery_queue
+ADD COLUMN delivery_response_time INTEGER;
```
alexbishop1, line 2: Maybe indicate the units in the name (seconds, milliseconds etc.)?

alexbishop1 (approved):

PR #273, opened 2022-06-10
BAU Improve logging 

sfount:
For consistency with other apps, add the commons `LoggingFilter` for
structured request logs.

Add a custom request filter to take webhook context specific vars from
the path and query params and provide them to all subsequent logs.

Add structured logs to the two main webhook processes 
* recording messages to be sent by consuming the event stream
* sending webhook messages

----

Refactor code for sending messages to have a single method to handle
responses. This allows us to uniformly add a structured log that will be
easy to grep/ process in Splunk.

---

Tests should still pass.


stephencdaly (approved):
Looks all good!

PR #297, opened 2022-06-22
BAU Filter out events that don't contain service id or live properties

sfount:
The deployed event message handler is failing to process some messages
emitted by connector that are missing service id and live properties.
This could eventually be a pact contract requirement for system events
however for now the process should be able to manage this case (in
favour of NPE).

Check for these properties before using them to check for webhook
subscriptions with the DAO.

Add coverage to check that events missing these properties are filtered
out before being handed to the webhook message processing code.

Change the event handling rescheduling log from an error to a warn, this
error should be on events eventually making their way to the dead letter
queue.

### DIFF
--- a/src/test/java/uk/gov/pay/webhooks/queue/EventQueueTest.java
+++ b/src/test/java/uk/gov/pay/webhooks/queue/EventQueueTest.java
@@ -1,0 +1,89 @@
+package uk.gov.pay.webhooks.queue;
+
+import ch.qos.logback.classic.Logger;
+import ch.qos.logback.classic.spi.ILoggingEvent;
+import ch.qos.logback.classic.spi.LoggingEvent;
+import ch.qos.logback.core.Appender;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.ArgumentCaptor;
+import org.mockito.Captor;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+import org.slf4j.LoggerFactory;
+import uk.gov.pay.webhooks.app.SqsConfig;
+import uk.gov.pay.webhooks.app.WebhooksConfig;
+import uk.gov.pay.webhooks.queue.sqs.QueueException;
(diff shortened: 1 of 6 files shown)


src/test/java/uk/gov/pay/webhooks/queue/EventQueueTest.java
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    @Test
+    public void shouldFilterEventsWithoutValidPropertiesAndLog() throws QueueException {
+        Logger root = (Logger) LoggerFactory.getLogger(EventQueue.class);
+        root.addAppender(mockAppender);
+
+        var sqsMessageWithMissingServiceId = """
+                {
+                  "Message" : "{\\"sqs_message_id\\":\\"dc142884-1e4b-4e57-be93-111b692a4868\\",\\"live\\":false,\\"resource_type\\":\\"payment\\",\\"resource_external_id\\":\\"t8cj9v1lci7da7pbp99qg9olv3\\",\\"parent_resource_external_id\\":null,\\"timestamp\\":\\"2019-08-31T14:18:46.446541Z\\",\\"event_type\\":\\"PAYMENT_DETAILS_ENTERED\\",\\"reproject_domain_object\\":false}"
```
alexbishop1, line 61: This JSON in JSON makes my head hurt!

Maybe we could do something like this:

```java
var sqsMessageWithMissingServiceIdEnclosedMessage = """
        {
          "sqs_message_id": "dc142884-1e4b-4e57-be93-111b692a4868",
          "live": false,
          "resource_type": "payment",
          "resource_external_id": "t8cj9v1lci7da7pbp99qg9olv3",
          "parent_resource_external_id": null,
          "timestamp": "2019-08-31T14:18:46.446541Z",
          "event_type": "PAYMENT_DETAILS_ENTERED",
          "reproject_domain_object": false
        }
        """;
        
var sqsMessageWithMissingServiceId = """
        {
          "Message" : "%s"
        }
        """.formatted(StringEscapeUtils.escapeJson(sqsMessageWithMissingServiceIdEnclosedMessage));
```

(Whitespace will be different but that should not matter.)

alexbishop1 (changes requested):
Looks good, just one thing about the tests.

alexbishop1 (dismissed):

alexbishop1 (approved):

PR #308, opened 2022-06-30
BAU Include live and event type properties on logs when working on a webhook message

sfount:
When doing work on a specific webhook message, include the live and
event type properties.

These can be used to filter logs in queries.


stephencdaly (approved):
LGTM!

PR #362, opened 2022-08-08
PP-9016 Check domain is in allow list before sending message

sfount:
Check that the domain of a webhook callback URL is in the allow
list before sending a webhook message.

The allow list is a simple list of domains like gov.uk and
example.com. When sending a message, we check if the domain is in
the allow list. Subdomains are allowed, so if the allow list
contains gov.uk and example.com, we will allow callback URLs with
domains like www.service.gov.uk and www.example.com. Since domain
names are case-insensitive, so is the check (for plain ASCII
domain names).

The implementation relies on Guava’s InternetDomainName class.
Each domain name on the allow list is converted into an
InternetDomainName. We then convert the callback URL to a URL
object (we chose this over an URI object because URL is more
specific for web addresses and supports international domain
names) and extract the hostname. If the URL has a host name
and it’s a domain name (rather than an IP address), we convert
it into an InternetDomainName (if it neither of these things
are true, we disallow the sending of the webhook message). We then
check if the InternetDomainName in is in the allow list. If it is,
we will allow the webhook message to be sent. If it is not, we lop
the leftmost part of the domain name off (for example,
www.service.gov.uk becomes service.gov.uk) to create a new
InternetDomainName and then look for that in the allow list. We
continue doing this until there are just two parts of the domain
left and do a final check on that (this means a single-part domain
name like com will never be found in the allow list). If we have
not found the domain name in the allow list by this point, we do
not allow the webhook message to be sent.

We made a choice to rely heavily on the InternetDomainName class
to avoid having to deal with domain name weirdness ourselves. An
earlier draft of this code kept the allow list as strings and
avoided creating a new InternetDomainName each time a part is
lopped off the domain name (with allow its associated parsing and
validation) but testing showed this was only roughly three times
faster, which did not seem a big enough win when considered against
the simplicity of just using InternetDomainName.

International domain names are supported by the check does not
consider IDNs and their Punycode equivalents to match. If we ever
have to add an IDN to the whitelist, we could add the Punycode
equivalent as well to ensure both work.

with @alexbishop1

### DIFF
--- a/src/main/java/uk/gov/pay/webhooks/validations/CallbackUrlService.java
+++ b/src/main/java/uk/gov/pay/webhooks/validations/CallbackUrlService.java
@@ -1,12 +1,11 @@
 package uk.gov.pay.webhooks.validations;
 
 import com.google.common.net.InternetDomainName;
-import org.eclipse.persistence.queries.Call;
+import org.apache.commons.lang3.StringUtils;
 import uk.gov.pay.webhooks.app.WebhooksConfig;
 
 import java.net.MalformedURLException;
 import java.net.URL;
-import java.util.Optional;
 import java.util.Set;
 
 import static java.util.stream.Collectors.toUnmodifiableSet;
@@ -33,27 +32,29 @@ public class CallbackUrlService {
             throw new CallbackUrlMalformedException("Callback URL is not a valid URL");
         }
 
-        if (url.getHost().isEmpty()) {
+        if (StringUtils.isEmpty(url.getHost())) {
             throw new CallbackUrlMalformedException("Callback URL must contain a host");
         }
 
-        if (!url.getProtocol().equals("https")) {
+        if (!"https".equals(url.getProtocol())) {
             throw new CallbackUrlProtocolNotSupported("Callback URL must use HTTPS protocol");
+        }
+
+        if (!InternetDomainName.isValid(url.getHost())) {
+            throw new CallbackUrlMalformedException("Callback URL host must be a domain name");
         }
 
         return url;
     }
 
     private void validateUrlIsInLiveDomains(URL callbackUrl) {
-        if (InternetDomainName.isValid(callbackUrl.getHost())) {
-            var domain = InternetDomainName.from(callbackUrl.getHost());
-            while (domain.hasParent()) {
-                if (allowedDomains.contains(domain)) {
-                    return;
-                }
-                domain = domain.parent();
+        var domain = InternetDomainName.from(callbackUrl.getHost());
+        while (domain.hasParent()) {
+            if (allowedDomains.contains(domain)) {
+                return;
             }
+            domain = domain.parent();
         }
-        throw new DomainNotOnAllowListException(callbackUrl.getHost() + " is not in the allow list");
+        throw new CallbackUrlDomainNotOnAllowListException(callbackUrl.getHost() + " is not in the allow list");
     }
 }
--- a/src/main/java/uk/gov/pay/webhooks/validations/CallbackUrlService.java
+++ b/src/main/java/uk/gov/pay/webhooks/validations/CallbackUrlService.java
@@ -1,0 +1,59 @@
+package uk.gov.pay.webhooks.validations;
+
+import com.google.common.net.InternetDomainName;
+import org.eclipse.persistence.queries.Call;
+import uk.gov.pay.webhooks.app.WebhooksConfig;
+
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.util.Optional;
+import java.util.Set;
+
+import static java.util.stream.Collectors.toUnmodifiableSet;
+
+public class CallbackUrlService {
+    private final Set<InternetDomainName> allowedDomains;
+
+    public CallbackUrlService(WebhooksConfig webhooksConfig) {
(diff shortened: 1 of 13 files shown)


src/main/java/uk/gov/pay/webhooks/validations/CallbackUrlService.java
```diff
@@ -0,0 +1,59 @@
...
+
+    private URL validateAndGetUrlIsWellFormed(String callbackUrl) throws MalformedURLException {
+        URL url;
+        try {
+            url = new URL(callbackUrl);
+        } catch (MalformedURLException e) {
+            throw new CallbackUrlMalformedException("Callback URL is not a valid URL");
+        }
+
+        if (url.getHost().isEmpty()) {
```
alexbishop1, line 36: We’re relying on the not-entirely-clearly-specified behaviour that `getHost()` returns `""` rather than `null` if there is no host.

src/main/java/uk/gov/pay/webhooks/validations/CallbackUrlService.java
```diff
@@ -0,0 +1,59 @@
...
+            throw new CallbackUrlMalformedException("Callback URL must contain a host");
+        }
+
+        if (!url.getProtocol().equals("https")) {
```
alexbishop1, line 40: I think protocols are case-insensitive — does some normalisation take care of this? Perhaps a test to make sure `HTTPS` is allowed would be useful.

src/main/java/uk/gov/pay/webhooks/validations/CallbackUrlService.java
```diff
@@ -0,0 +1,59 @@
...
+            throw new CallbackUrlProtocolNotSupported("Callback URL must use HTTPS protocol");
+        }
+
+        return url;
+    }
+
+    private void validateUrlIsInLiveDomains(URL callbackUrl) {
+        if (InternetDomainName.isValid(callbackUrl.getHost())) {
```
alexbishop1, line 48: This is where we’re checking the host is not an IP address. I wonder if an explicit check in `validateAndGetUrlIsWellFormed(String)` might be clearer.

src/main/java/uk/gov/pay/webhooks/validations/DomainNotOnAllowListException.java
```diff
@@ -0,0 +1,9 @@
+package uk.gov.pay.webhooks.validations;
+
+public class DomainNotOnAllowListException extends RuntimeException {
```
alexbishop1, line 3: Maybe this should be renamed `CallbackUrlDomainNotOnAllowListException` to match the other exceptions.

alexbishop1 (changes requested):
I very much like that `CallbackUrlService` is now its own class. One comment about checking whether the protocol can be case-insensitive and another about the naming of an exception.

src/main/java/uk/gov/pay/webhooks/validations/CallbackUrlService.java
sfount: 🤔 yeah maybe we just remove this, I'm fairly confident the URL won't be instantiated if there's no host but we could double check that behaviour

src/main/java/uk/gov/pay/webhooks/validations/CallbackUrlService.java
sfount: It is indeed normalised by the java net URL, I'll add an assertion for this

src/main/java/uk/gov/pay/webhooks/validations/CallbackUrlService.java
alexbishop1: No, I think we need it. Hostless URLs are a thing. For example, `file:///etc/etc` has no host (but it does have a protocol and a path etc.).

src/main/java/uk/gov/pay/webhooks/validations/CallbackUrlService.java
```diff
@@ -0,0 +1,60 @@
...
+
+    private URL validateAndGetUrlIsWellFormed(String callbackUrl) throws MalformedURLException {
+        URL url;
+        try {
+            url = new URL(callbackUrl);
+        } catch (MalformedURLException e) {
+            throw new CallbackUrlMalformedException("Callback URL is not a valid URL");
+        }
+
+        if (Optional.ofNullable(url.getHost()).map(String::isEmpty).orElse(true)) {
```
alexbishop1, line 35: `StringUtils.isEmpty(url.getHost())` would also work, of course.

src/main/java/uk/gov/pay/webhooks/validations/CallbackUrlService.java
```diff
@@ -0,0 +1,60 @@
...
+            throw new CallbackUrlMalformedException("Callback URL must contain a host");
+        }
+
+        if (!"https".equals(url.getProtocol())) {
+            throw new CallbackUrlProtocolNotSupported("Callback URL must use HTTPS protocol");
+        }
+
+        if (!InternetDomainName.isValid(url.getHost())) {
+            throw new CallbackUrlMalformedException("Callback URL is not a valid domain name");
```
alexbishop1, line 44: Really nitty but I think something like “Callback URL host must be a domain name” might be best since if we hit this it’s most likely that we have an IP address (an invalid domain name should not get this far).

alexbishop1 (approved):

alexbishop1 (approved):

PR #366, opened 2022-08-09
PP-9016 Hook callback URL validation into create and update

sfount:
Follows on from https://github.com/alphagov/pay-webhooks/pull/362

Include a call to the the callback url service on the create and update
webhook resource to ensure the url is validated whenever a webhook is
mutated. This requires passing down the live context for any given
webhook resource.

There's a remaining developer note to demonstrate that we could move the
create webhook validation to a custom constraint validator to bring it
inline with other basic validation done for create requests. Currently
this is entirely separate from the validation done against the json
patch logic (which is something else that could be improved in the
future).

For now this will throw separate and one off exceptions for the callback
url on the update endpoint, rather than throwing a `ValidationException`
to be caught by the commons code and wrapped into a list of exceptions
to be returned at once. This allows the `error_identifier` to be simply
returned to the frontend without having consider how this would be
_optionally_ represented and nested nested in the java commons code.
Further work could refine this to support nested and batch error
identifiers however currently the frontend will only be leaning on the
backend webhooks resource to do business logic validation around the
URL provided.

Splits the json patch validator into two depending on the live context
of the webhook, this feels like a good compromise between instantiating
the objects once but allowing the validation to be live context
dependant.

### DIFF
--- a/src/main/java/uk/gov/pay/webhooks/validations/WebhookRequestValidator.java
+++ b/src/main/java/uk/gov/pay/webhooks/validations/WebhookRequestValidator.java
@@ -1,13 +1,14 @@
 package uk.gov.pay.webhooks.validations;
 
 import com.fasterxml.jackson.databind.JsonNode;
+import com.google.inject.Inject;
+import uk.gov.pay.webhooks.webhook.resource.CreateWebhookRequest;
 import uk.gov.service.payments.commons.api.validation.JsonPatchRequestValidator;
 import uk.gov.service.payments.commons.api.validation.PatchPathOperation;
 import uk.gov.service.payments.commons.model.jsonpatch.JsonPatchOp;
 import uk.gov.service.payments.commons.model.jsonpatch.JsonPatchRequest;
 
 import java.util.Map;
-import java.util.function.Consumer;
 
 import static uk.gov.pay.webhooks.webhook.resource.WebhookResponse.FIELD_CALLBACK_URL;
 import static uk.gov.pay.webhooks.webhook.resource.WebhookResponse.FIELD_DESCRIPTION;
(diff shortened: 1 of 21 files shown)


src/main/java/uk/gov/pay/webhooks/validations/WebhookRequestValidator.java
```diff
@@ -16,16 +17,44 @@
...
+
+    private final JsonPatchRequestValidator liveContextValidator = validator(true);
+    private final JsonPatchRequestValidator testContextValidator = validator(false);
+
+    @Inject
+    public WebhookRequestValidator(CallbackUrlService callbackUrlService) {
+        this.callbackUrlService = callbackUrlService;
+    }
+
+    public void validate(JsonNode payload, Boolean isLiveContext)  {
```
alexbishop1, line 30: `isLiveContext` is a boolean acting as a poor man’s enum. I think I’d prefer an actual enum, particularly to make `WebhookRequestValidatorTest` nicer.

src/main/java/uk/gov/pay/webhooks/webhook/exception/WebhooksException.java
```diff
@@ -0,0 +1,23 @@
+package uk.gov.pay.webhooks.webhook.exception;
+
+public class WebhooksException extends RuntimeException {
+    protected static final String IDENTIFIER_GENERIC = "generic";
+    protected static final String IDENTIFIER_CALLBACK_URL_NOT_ON_ALLOW_LIST = "callback_url_not_on_allow_list";
+    protected static final String IDENTIFIER_CALLBACK_URL_MALFORMED = "callback_url_malformed";
+    protected static final String IDENTIFIER_CALLBACK_URL_PROTOCOL_NOT_SUPPORTED = "callback_url_protocol_not_supported";
+
+    private String errorIdentifier = IDENTIFIER_GENERIC;
```
alexbishop1, line 9: This should really be an enum. I know that would conflict with the `ErrorIdentifier` defined in Java commons but I think having an `ErrorIdentifierForCallbackUrlErrors` or `ErrorIdentifierForCallbackUrlErrorsIPinkySwearWeWillMergeThisWithTheMainErrorIdentifierSomeDay` would be preferable to using strings.

alexbishop1 (changes requested):
Looks fine but I want more enums!

Do we have a test that explicitly checks that the allow list is not used for test webhooks?

src/main/java/uk/gov/pay/webhooks/webhook/exception/WebhooksException.java
sfount: It's only the admin tool that consumes the resources that can throw these exceptions at preset. 

I don't think we've got precedence for shared enums across frontend/ backend. 

Happy to help think it through but this enum currently couldn't be used by the consumer. I think it's a good idea to store them as one anyway to have one place to look in the Java app

src/main/java/uk/gov/pay/webhooks/validations/WebhookRequestValidator.java
sfount: 👍 I'm in favour of this idea but not quite sure how it would work with the other boolean-y live checks throughout the app. I suppose this could set precedent for building the enum from a webhook entity and having it currently specify `TEST` and `LIVE` values. That could then be used by the validator and services further down

alexbishop1 (approved):

src/test/java/uk/gov/pay/webhooks/webhook/resource/WebhookResourceIT.java
```diff
@@ -65,13 +57,41 @@ public void shouldCreateAndRetrieveAWebhook() {
...
+    @Test
+    public void shouldCreateRejectWebhookForKnownErrorIdentifiers() {
+        given()
+                .port(port)
+                .contentType(JSON)
+                .body(createWebhookRequestBody("http://gov.uk", true))
+                .post("/v1/webhook")
+                .then()
+                .statusCode(Response.Status.BAD_REQUEST.getStatusCode())
+                .body("error_identifier", is("CALLBACK_URL_PROTOCOL_NOT_SUPPORTED"));
```
alexbishop1, line 76: `CALLBACK_URL_PROTOCOL_NOT_SUPPORTED.toString()` would work as well, of course.

alexbishop1 (approved):

PR #371, opened 2022-08-10
PP-9016 Check allow list at point of sending a message

sfount:
Adds `WONT_SEND` status for message delivery attempts.

For live messages only, validates that the callback url being delivered
to at the time is in the allow list.

Adds logging expected by ticket for tying into protective monitoring.

### DIFF
--- a/src/main/java/uk/gov/pay/webhooks/deliveryqueue/dao/WebhookDeliveryQueueEntity.java
+++ b/src/main/java/uk/gov/pay/webhooks/deliveryqueue/dao/WebhookDeliveryQueueEntity.java
@@ -76,7 +76,7 @@ public class WebhookDeliveryQueueEntity {
         PENDING,
         SUCCESSFUL,
         FAILED,
-        WONT_SEND
+        WILL_NOT_SEND
     }
 
 
--- a/src/main/java/uk/gov/pay/webhooks/deliveryqueue/dao/WebhookDeliveryQueueEntity.java
+++ b/src/main/java/uk/gov/pay/webhooks/deliveryqueue/dao/WebhookDeliveryQueueEntity.java
@@ -75,7 +75,8 @@ public class WebhookDeliveryQueueEntity {
     public enum DeliveryStatus {
         PENDING,
         SUCCESSFUL,
-        FAILED
+        FAILED,
+        WONT_SEND
     }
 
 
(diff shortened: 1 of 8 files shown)


src/main/java/uk/gov/pay/webhooks/deliveryqueue/dao/WebhookDeliveryQueueEntity.java
```diff
@@ -75,7 +75,8 @@ public Instant getCreatedDate() {
     public enum DeliveryStatus {
         PENDING,
         SUCCESSFUL,
-        FAILED
+        FAILED,
+        WONT_SEND
```
alexbishop1, line 79: I wonder if `WILL_NOT_SEND` might get around the inability to include apostrophes (compounded by the fact that ‘wont’ is an actual word).

src/main/java/uk/gov/pay/webhooks/deliveryqueue/managed/SendAttempter.java
```diff
@@ -53,13 +61,13 @@ public SendAttempter(WebhookDeliveryQueueDao webhookDeliveryQueueDao,
...
     public void attemptSend(WebhookDeliveryQueueEntity queueItem) {
+        var webhook = queueItem.getWebhookMessageEntity().getWebhookEntity();
         var retryCount = webhookDeliveryQueueDao.countFailed(queueItem.getWebhookMessageEntity());
         Instant start = instantSource.instant();
 
         try {
-            var url = queueItem.getWebhookMessageEntity().getWebhookEntity().getCallbackUrl();
-            var uri = new URI(url);
-            
+            var uri = new URI(webhook.getCallbackUrl());
```
alexbishop1, line 69: We’re using `URI` rather than `URL` here, the latter being what we use in the validation. We should probably standardise on `URL`.

alexbishop1 (commented):
Looks good so far.

What do you think about having `getCallbackUrl()` on `Webhook` return a `URL` rather than a `String`? It would mean that whoever’s calling the method can be confident they have an actual URL (and not just a string that might be one). We should not need an error-checking for the `String`-to-`URL` conversion because anything persisted should be a valid URL already. This would mean we’d need to tweak the `CallbackUrlService` so that it can take a `URL` to validate rather than just `String`s.

alexbishop1 (approved):
Agreed that we’ll look at making `getCallbackUrl()` on `WebhookEntity` return a `URL` rather than a `String` in another PR.

PR #373, opened 2022-08-11
BAU — Set hibernate.generate_statistics=false

alexbishop1:
We do not think we need Hibernate statistics for now.


alexbishop1:
This did not seem to reduce the logging like we expected.

PR #381, opened 2022-08-15
PP-9685 Dont process events for disabled webhook

sfount:
Depends on https://github.com/alphagov/pay-webhooks/pull/380

When a webhook has status `DISABLED` do not include it in the valid
webhook subscriptions.

Add additional coverage to assert a webhook can be moved into this
state by consumers.


katstevens:
@sfount can you remember if this was done elsewhere or not? Can we close this PR?

SandorArpa:
Story's been played, no longer needed

PR #387, opened 2022-08-18
PP-9274 Update internal hooks for subscriptions

sfount:
Depends on https://github.com/alphagov/pay-webhooks/pull/386

Introduces new failed and expired mappings to internal events.

Update successful event mappings, previous "authorisation success" had
been used however this event is reached before a paying user "confirms"
a payment and the payment can still be cancelled at this point in time.
Instead of this use the confirmed for capture event which will guarantee
that the payment will be taken.

### DIFF
--- a/src/main/java/uk/gov/pay/webhooks/message/EventMapper.java
+++ b/src/main/java/uk/gov/pay/webhooks/message/EventMapper.java
@@ -10,21 +10,30 @@ import java.util.Optional;
 import java.util.stream.Collectors;
 
 public class EventMapper {
-    private static final Map<String, EventTypeName> internalToWebhook = Map.of(
-            "PAYMENT_STARTED", EventTypeName.CARD_PAYMENT_STARTED,
-            "AUTHORISATION_SUCCEEDED", EventTypeName.CARD_PAYMENT_SUCCEEDED,
-            "CAPTURE_CONFIRMED", EventTypeName.CARD_PAYMENT_CAPTURED,
-            "REFUND_SUCCEEDED", EventTypeName.CARD_PAYMENT_REFUNDED
+    private static final Map<String, EventTypeName> internalToWebhook = Map.ofEntries(
+            Map.entry("USER_APPROVED_FOR_CAPTURE", EventTypeName.CARD_PAYMENT_SUCCEEDED),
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/pay/webhooks/message/EventMapper.java
```diff
@@ -10,21 +10,30 @@
...
+            Map.entry("CANCELLED_BY_EXTERNAL_SERVICE", EventTypeName.CARD_PAYMENT_FAILED),
+            Map.entry("CANCELLED_BY_USER", EventTypeName.CARD_PAYMENT_FAILED),
+            Map.entry("CANCELLED_BY_EXPIRATION", EventTypeName.CARD_PAYMENT_EXPIRED),
+            Map.entry("CAPTURE_CONFIRMED", EventTypeName.CARD_PAYMENT_CAPTURED),
+            Map.entry("REFUND_SUCCEEDED", EventTypeName.CARD_PAYMENT_REFUNDED)
     );
 
-    private static final Map<EventTypeName, String> webhookToInternal = internalToWebhook
-            .entrySet()
+    private static final Map<EventTypeName, List<String>> webhookToInternal = internalToWebhook
```
alexbishop1, line 29: This looks like a constant, which should be `SCREAMING_SNAKE_CASE` 😱🐍

src/main/java/uk/gov/pay/webhooks/message/EventMapper.java
```diff
@@ -10,21 +10,30 @@
...
+            .keySet()
             .stream()
-            .collect(Collectors.toMap(Map.Entry::getValue, Map.Entry::getKey));
+            .collect(Collectors.groupingBy(internalToWebhook::get));
 
     private static final List<EventTypeName> childEvents = List.of(EventTypeName.CARD_PAYMENT_REFUNDED);
 
-    public static Optional<String> getInternalEventNameFor(EventTypeName webhookEventTypeName) {
+    public static Optional<List<String>> getInternalEventNameFor(EventTypeName webhookEventTypeName) {
```
alexbishop1, line 36: Two things:

1. Could we return an empty list when there’s nothing rather than an optional?
2. Can this be a set rather than a list?

alexbishop1 (changes requested):
Looks fine. Just a couple of mostly stylistic things.

alexbishop1 (approved):

PR #403, opened 2022-08-30
PP-9016 Compare apache and java net http client

sfount:
The apache client gives us more control over:
- connection pool behaviour, specifically around timeout behaviours for
  connection pools where we have no control over the receiving servers
  connection behaviour. We suspect these connection pools being
  terminated result in intermittent "Connection Reset" errors
- ease of specifying TLS and cipher requirements

### DIFF
--- a/src/main/java/uk/gov/pay/webhooks/app/WebhooksModule.java
+++ b/src/main/java/uk/gov/pay/webhooks/app/WebhooksModule.java
@@ -10,13 +10,10 @@ import com.google.inject.Provides;
 import io.dropwizard.hibernate.HibernateBundle;
 import io.dropwizard.hibernate.UnitOfWorkAwareProxyFactory;
 import io.dropwizard.setup.Environment;
-import org.apache.http.client.HttpClient;
 import org.apache.http.client.config.RequestConfig;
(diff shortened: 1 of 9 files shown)


src/main/java/uk/gov/pay/webhooks/app/WebhooksModule.java
```diff
@@ -54,8 +56,33 @@ public IdGenerator externalIdGenerator() {
...
+        var timeoutInMillis = Math.toIntExact(configuration.getQueueMessageReceiverConfig().getRequestTimeout().toMilliseconds());
+        var config = RequestConfig.custom()
+                .setConnectTimeout(timeoutInMillis)
+                .setConnectionRequestTimeout(timeoutInMillis)
+                .setSocketTimeout(timeoutInMillis)
+                .build();
+
+        var sslsf = new SSLConnectionSocketFactory(
+                SSLContexts.createDefault(),
+                new String[] { "TLSv1.2", "TLSv1.3" },
```
alexbishop1, line 75: Someone should tell the `SSLConnectionSocketFactory` developers about enums…

alexbishop1 (approved):
Looks fine and since it’s not really live yet there will never be a better chance to try it.

PR #406, opened 2022-08-31
BAU fix ignoring SQS event messages not marked as complete

sfount:
Currently events from the event stream are ignored when they're missing
properties required to check for webhook subscriptions (live and
service_id).

This change moves the logic for not handling the event into the webhook
message service handler in favour of the code receiving and
deserialising the event. This will mean that when the handler doesn't
throw any exceptions, the event will be marked as processed (and
ignored).


katstevens (approved):
This does look a lot simpler!

PR #407, opened 2022-08-31
BAU Disable hibernate statistics logs

sfount:
Hibernate seems to be generating and logging statistics for every single
database connection, by default.

How this behaviour was introduced or is on by default isn't clear at
this stage, however with production volumes the logs generated are
problematic.

Disable these logs by setting the hibernate property and investigate
why this behaviour is enabled in the first place.

Relates to https://github.com/alphagov/pay-webhooks/pull/373

### DIFF
--- a/src/main/resources/config/config.yaml
+++ b/src/main/resources/config/config.yaml
@@ -33,6 +33,7 @@ database:
   properties:
     charSet: UTF-8
     hibernate.dialect: org.hibernate.dialect.PostgreSQLDialect
+    hibernate.generate_statistics: false
 
   # the maximum amount of time to wait on an empty pool before throwing an exception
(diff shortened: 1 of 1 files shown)


src/main/resources/config/config.yaml
```diff
@@ -33,6 +33,7 @@ database:
   properties:
     charSet: UTF-8
     hibernate.dialect: org.hibernate.dialect.PostgreSQLDialect
+    hibernate.generate_statistics: false
```
alexbishop1, line 36: So this is where you’re supposed to put it?

(Previous attempt: https://github.com/alphagov/pay-webhooks/pull/373)

alexbishop1 (approved):

PR #408, opened 2022-08-31
PP-9016 Fix webhook sender timeout config

sfount:
The configuration properties for message sending client timeouts were
applied to the wrong configuration file.

This passed tests because the `test-config.yaml` correctly (incorrectly)
set the configuration on the incorrect property.

The real configuration file `config.yaml` incorrectly (correctly) set
the configuration on the right property however this couldn't be parsed
by the app causing it not to start.


katstevens (approved):
Easily done!

PR #560, opened 2022-12-08
PP-10191 Handle unknown exceptions on the message to send polling thread

sfount:
Currently if the webhook delivery queue DAO `nextToSend` fails to connect to the database, an exception will be raised that won't be handled by the polling wrapper code. This will cause the scheduled executor service the code is running in to stop and it has no way to recover other than an app restart.

In any other code we have the runs managed background tasks, there's generally a top level exception handler to log when something has gone completely wrong (i.e couldn't access a database during an short upgrade window) and allow the thread to continue safely.

Add the exception handling in the main `sendIfAvailable` wrapper as this is where we can most clearly clean up the MDC logging contexts, none of the thread code that runs outside of this should be able to raise exceptions. This handler covers any interaction with the database or REST clients for attempted message sending (which handles its own errors separately).


SandorArpa (approved):
LGTM

PR #564, opened 2022-12-12
PP-10433 Fix webhooks message poller thread configuration

sfount:
Address two problems in the scheduled thread executor configuration:

- passing in a service method by reference doesn't seem to work when the thread tries to restart. Following examples elsewhere passing in something in the method context (`this::process`) seems to work
- although there is a configuration for number of threads for the message sender, it only ever actually spins up one thread (although allocating the space for the total number of threads). Following examples elsewhere, iterate through the number of threads and spin them up -- include the thread number in the thread name

There should never really only be 1 thread running for either background process. Set sensible defaults which can be tweaked with performance testing.


SandorArpa (approved):
LGTM

PR #615, opened 2023-01-17
PP-10539: Use standard cookie spec

oswaldquek:
Sometimes the following is logged when making a call to a webhook:
```
   logger_name: org.apache.http.client.protocol.ResponseProcessCookies
   message: Invalid cookie header: "Set-Cookie: session.sig=5LqAuxeg8FDwdDJ4xCjTGG92AVU; path=/; expires=Tue, 17 Jan 2023 10:31:32 GMT; httponly". Invalid 'expires' attribute: Tue, 17 Jan 2023 10:31:32 GMT
```

This log appears in conjunction with webhooks keeping the http connection open to webhooks egress for 15 minutes. While it isn't clear why this would be the case, and even if this commit doesn't resolve the issue, it's probably a good idea to make the http client understand the latest RFC-compliant headers (see https://stackoverflow.com/questions/36473478/fixing-httpclient-warning-invalid-expires-attribute-using-fluent-api for reference).


oswaldquek:
Closing as this PR didn't achieve the supposed effect.

sfount (approved):
We've observed this resolve (logged) issues with response cookie parsing in the test environment, change looks good. We can review if this behaviour is expected as service developers integrate with webhooks

PR #618, opened 2023-01-18
bau: Unblock failed GHA

oswaldquek:
Apparently GH workflows don't run with
actions/setup-java@1df8dbefe2a8cbc99770194893dd902763bee34b. Version c3ac5dd0ed8db40fedb61c32fbe677e6b355e94c is in our [whitelist](https://github.com/alphagov/pay-webhooks/settings/actions) so should be ok.

### DIFF
--- a/.github/workflows/post-merge.yml
+++ b/.github/workflows/post-merge.yml
@@ -26,7 +26,7 @@ jobs:
       - name: Git checkout
         uses: actions/checkout@755da8c3cf115ac066823e79a1e1788f8940201b
       - name: Set up JDK 17
-        uses: actions/setup-java@a18c333f3f14249953dab3e186e5e21bf3390f1d
+        uses: actions/setup-java@c3ac5dd0ed8db40fedb61c32fbe677e6b355e94c
         with:
           java-version: '17'
           distribution: 'adopt'


katstevens:
What version of setup-java is this? Will we need to adjust the dependabot config?

oswaldquek:
> What version of setup-java is this? Will we need to adjust the dependabot config?

What do you mean? It's version c3ac5dd0ed8db40fedb61c32fbe677e6b355e94c. Not sure if we'll need to adjust the dependabot config.

katstevens:
> > What version of setup-java is this? Will we need to adjust the dependabot config?
> 
> What do you mean? It's version c3ac5dd0ed8db40fedb61c32fbe677e6b355e94c. Not sure if we'll need to adjust the dependabot config.

I meant which semantic version does that hash refer to (e.g. v3.9.0 is `1df8dbefe2a8cbc99770194893dd902763bee34b` on [this dependabot PR](https://github.com/alphagov/pay-webhooks/pull/569/files)). If we merge this, then dependabot will try and update to the latest version. We can pin the dependabot config to stop it doing that if needed.

oswaldquek:
Ah, the commit c3ac5dd0ed8db40fedb61c32fbe677e6b355e94c can be found at https://github.com/actions/setup-java/compare/v3.6.0...actions:v3.8.0 so the version would be 3.8.0.

oswaldquek:
Do we want one of these in the dependabot config? https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuring-dependabot-version-updates#example-disabling-version-updates-for-some-dependencies
I think we probably want 
```
      - dependency-name: "actions/setup-java"
        # Ignore only new versions for 4.x and 5.x
        versions: ["3.x"]
```

katstevens:
I'm a little confused as to why 3.9.0 doesn't work here - [it was fine for connector](https://github.com/alphagov/pay-connector/pull/4207) so it would be good to know why.

oswaldquek:
> I'm a little confused as to why 3.9.0 doesn't work here - [it was fine for connector](https://github.com/alphagov/pay-connector/pull/4207) so it would be good to know why. I've rebased [the other PR](https://github.com/alphagov/pay-webhooks/pull/569) in case it was just a stale branch.

It'd be good to know why too, but I stopped short of investigating as this is [supposedly a known issue](https://gds.slack.com/archives/CAD6G150R/p1674041206860119?thread_ts=1674040818.691329&cid=CAD6G150R).

oswaldquek:
Closing as the issue is resolved: https://github.com/alphagov/pay-webhooks/actions/runs/3957817175.

PR #625, opened 2023-01-23
PP-10539: Set keep alive strategy

oswaldquek:
According to the HttpClient 4.3.3. reference: “If the Keep-Alive header is not
present in the response, HttpClient assumes the connection can be kept alive
indefinitely.” Although this change didn't solve the issue where
webhooks-egress logs `CONNECT gds-club-milk-demo.cloudapps.digital:443` 15
minutes after the webhook was called (see
https://payments-platform.atlassian.net/browse/PP-10539 for more details), it
is probably a good idea to ensure connections are dropped after 5 seconds here.


katstevens:
@oswaldquek what's the current status of this PR? Do we need input on particular people to move this along?

oswaldquek:
This PR was manually deployed but it didn't fix the issue (see my comment on 23 January 2023 at 11:24 in the JIRA story).

PR #626, opened 2023-01-23
PP-10631: Specify URL when logging 'Sending webhook message finished'

oswaldquek:
As part of protective monitoring, there is an alert to compare webhooks and webhooks-egress target domains. Currently the alert's search on the webhooks microservice logs is searching for "Sending webhook message". However this isn't correct as the presence of this log doesn't indicate webhooks microservice actually did call out to the target url. This is because at the point of this log, the target webhook url may be inactive, not on the allow list or failed because of some HTTP failure (e.g. socket timeout, interrupted exception). The splunk alert should search where the log has "Sending webhook message finished" and "to_state" is "SUCCESSFUL".

This commit adds the "domain" field to where "Sending webhook message finished" is logged so splunk can extract the log properly.

### DIFF
--- a/src/main/java/uk/gov/pay/webhooks/deliveryqueue/managed/SendAttempter.java
+++ b/src/main/java/uk/gov/pay/webhooks/deliveryqueue/managed/SendAttempter.java
@@ -16,7 +16,6 @@ import uk.gov.pay.webhooks.validations.CallbackUrlDomainNotOnAllowListException;
 import javax.inject.Inject;
 import javax.ws.rs.core.Response;
 import java.io.IOException;
-import java.net.MalformedURLException;
 import java.net.SocketTimeoutException;
 import java.net.URL;
 import java.net.http.HttpTimeoutException;
@@ -69,15 +68,8 @@ public class SendAttempter {
         var retryCount = webhookDeliveryQueueDao.countFailed(queueItem.getWebhookMessageEntity());
         Instant start = instantSource.instant();
 
-        URL url = null;
         try {
-            url = new URL(webhook.getCallbackUrl());
-        } catch (MalformedURLException e) {
-            LOGGER.error("Callback URL {} is malformed.", webhook.getCallbackUrl());
-            throw new RuntimeException(e);
-        }
-
-        try {
+            var url = new URL(webhook.getCallbackUrl());
 
             LOGGER.info(
                     Markers.append(WEBHOOK_CALLBACK_URL, queueItem.getWebhookMessageEntity().getWebhookEntity().getCallbackUrl())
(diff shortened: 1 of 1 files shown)


src/main/java/uk/gov/pay/webhooks/deliveryqueue/managed/SendAttempter.java
```diff
@@ -70,8 +72,15 @@ public void attemptSend(WebhookDeliveryQueueEntity queueItem) {
         var retryCount = webhookDeliveryQueueDao.countFailed(queueItem.getWebhookMessageEntity());
         Instant start = instantSource.instant();
 
+        URL url = null;
+        try {
+            url = new URL(webhook.getCallbackUrl());
+        } catch (MalformedURLException e) {
+            LOGGER.info(format("Callback URL %s is malformed.", webhook.getCallbackUrl()));
```
nlsteers, line 79:
```suggestion
            LOGGER.info("Callback URL {} is malformed.", webhook.getCallbackUrl());
```
You don't need the string format here, SLF4J can handle the param substitution

src/main/java/uk/gov/pay/webhooks/deliveryqueue/managed/SendAttempter.java
```diff
@@ -32,6 +33,7 @@
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
+import static java.lang.String.format;
```
nlsteers, line 36:
```suggestion
```

src/main/java/uk/gov/pay/webhooks/deliveryqueue/managed/SendAttempter.java
nlsteers, line 79: Should this be a `WARN`|`ERROR` rather than `INFO`?

src/main/java/uk/gov/pay/webhooks/deliveryqueue/managed/SendAttempter.java
```diff
@@ -70,8 +72,15 @@ public void attemptSend(WebhookDeliveryQueueEntity queueItem) {
         var retryCount = webhookDeliveryQueueDao.countFailed(queueItem.getWebhookMessageEntity());
         Instant start = instantSource.instant();
 
+        URL url = null;
```
nlsteers, line 75: Can this not be added to the existing try-catch block with a new catch block to cover the `MalformedURLException` handler? We wouldn't need to explicitly define url as null

oswaldquek: We can't add it to the existing try-catch block as the `url` variable won't be accessible by code in the other catch blocks where it's needed.

src/main/java/uk/gov/pay/webhooks/deliveryqueue/managed/SendAttempter.java
oswaldquek: I suppose ERROR is appropriate. There is validation on the selfservice UI though so this really shouldn't happen.

nlsteers (approved):
LGTM

PR #677, opened 2023-03-01
PP-10543 Add last delivery status to webhook messages

sfount:
Add a column to webhook messages to allow for simple filtering of messages by status in the admin tool. This column will line up with the existing status represented on delivery queue entries and should eventually be used to filter and display the delivery status for a given message.

### DIFF
--- a/src/main/resources/migrations/0012_alter_table_webhook_messages_add_last_delivery_status.sql
+++ b/src/main/resources/migrations/0012_alter_table_webhook_messages_add_last_delivery_status.sql
@@ -1,4 +1,8 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:alter-table-webhook-messages-add-last-delivery-status
 ALTER TABLE webhook_messages
 ADD COLUMN last_delivery_status VARCHAR(64) check (last_delivery_status in ('SUCCESSFUL', 'FAILED', 'PENDING', 'WILL_NOT_SEND'));
 
-CREATE INDEX IF NOT EXISTS last_delivery_status_idx on webhook_messages(last_delivery_status)
+--changeset uk.gov.pay:alter-table-webhook-messages-index-last-delivery-status runInTransaction:false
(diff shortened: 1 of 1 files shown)


src/main/resources/migrations/0012_alter_table_webhook_messages_add_last_delivery_status.sql
```diff
@@ -0,0 +1,4 @@
+ALTER TABLE webhook_messages
+ADD COLUMN last_delivery_status VARCHAR(64) check (last_delivery_status in ('SUCCESSFUL', 'FAILED', 'PENDING', 'WILL_NOT_SEND'));
+
+CREATE INDEX IF NOT EXISTS last_delivery_status_idx on webhook_messages(last_delivery_status)
```
kbottla, line 4: Should we create it `concurrently`?

sfount: I think we'd be OK to not create it concurrently as the column will be definitionally empty (having just been created if this migration is running) but happy to add it as best practice

src/main/resources/migrations/0012_alter_table_webhook_messages_add_last_delivery_status.sql
```diff
@@ -0,0 +1,4 @@
+ALTER TABLE webhook_messages
+ADD COLUMN last_delivery_status VARCHAR(64) check (last_delivery_status in ('SUCCESSFUL', 'FAILED', 'PENDING', 'WILL_NOT_SEND'));
```
kbottla, line 2: Do we need to check the list of values here? Adding a new enum value to the code could cause issues if there was no equivalent db migration. or at least we should have a test to check that db inserts still work

sfount: https://github.com/alphagov/pay-webhooks/commit/4b6553b8dc2c934d67ee96b503bf394faa653f0e

kbottla (approved):

PR #678, opened 2023-03-01
PP-10543 Validate delivery status enumerations against database schema

sfount:
Depends on https://github.com/alphagov/pay-webhooks/pull/677

Extract delivery status to be shared outside of the delivery queue class.

Ensure that any changes that are made to the enum representing possible
message delivery statuses are valid according to database constraints.
Do this by spinning up the database and running all of the enum values
against webhook message entries.


sfount:
Primary change in https://github.com/alphagov/pay-webhooks/pull/678/commits/4b6553b8dc2c934d67ee96b503bf394faa653f0e

kbottla (approved):

PR #680, opened 2023-03-01
BAU Favour enums and javax persistence (delivery queue)

sfount:
Instead of manually mapping between a string taken from the database, use the enumerated javax annotation to do this for us.


src/main/java/uk/gov/pay/webhooks/webhook/resource/WebhookResource.java
```diff
@@ -174,13 +174,10 @@ public List<WebhookResponse> getWebhooks(@Parameter(example = "true", descriptio
     public WebhookMessageSearchResponse getWebhookMessages(
             @Parameter(example = "gh0d0923jpsjdf0923jojlsfgkw3seg") @PathParam("webhookExternalId") String externalId,
             @Parameter(example = "1") @QueryParam("page") Integer page,
-            @Parameter(example = "SUCCESSFUL") @Valid @QueryParam("status") DeliveryStatus status
```
kbottla, line 177: The example here is still useful for API specs

sfount: Hm I didn't like that it wasn't related to the actual underlying enum and could easily get out of date, I can't quickly figure out how to reference the enum values though. Happy to put it back

kbottla (approved):

PR #707, opened 2023-03-15
PP-10543 Set webhook message last delivery status 

sfount:
When a webhook message is emitted, also record the delivery result on
the webhook message entity itself. This will be used for filtering and
aggregate stats in the future. This update is done within a single `@UnitOfWork` scope and given the database transaction guarantee that comes with that.

Instead of relying on the delivery queue status for the last attempt,
directly use the `lastDeliveryStatus` property now that it exists and is
kept up to date.

This will allow JPA to lazy load the last delivery queue which will
significatly decrease the complexity of the query for fetching the
messages based on status.

Tests for filtering/ paginating webhook messages should continue to
pass.

Some tests had to be moved around (outside of the app extension with background tasks enabled) which was done in https://github.com/alphagov/pay-webhooks/pull/707/commits/f31970c3a8b0a163b116dbafc1bc7f71ae56a7e6

---

Update the get one webhook message, list message and count messages query to directly reference the
webhook entity (using the internal identifier) rather than relying on a
sub query or a join.

Using postgres `explain` and `explain analyse` we've observed that using
the external ID from the right hand side of the join or adding a sub
query to lookup the webhook ID from an external ID both cause postgres
to not use appropriate indexes and do a full (reverse) table scan on the
webhook messages table.

As we don't need the full webhoook details in this instance, set the
fetch mode to lazy and directly reference the webhook entity to ensure
appropriate indexes are used for the query.

In order to do this cleanly, this commit also resolves an issue where
the previous get message DAO method was not returning an optional, which
meant an empty result would cause a null pointer exception when being
mapped to a JSON response. By making this optional at the DAO layer,
each subsequent layer (service then resource) handle the case of there
being no data appropriately. Add additional coverage to make sure we're
correctly mapping the get message resource to a 404.


src/test/java/uk/gov/pay/webhooks/message/dao/WebhookMessageDaoTest.java
kbottla, line 1: Not part of this PR but should this file name end with IT?

src/main/java/uk/gov/pay/webhooks/webhook/dao/WebhookDao.java
```diff
@@ -29,6 +29,14 @@ public Optional<WebhookEntity> findByExternalId(String webhookExternalId, String
                     .findFirst();
     }
 
+    public Optional<WebhookEntity> findByExternalId(String webhookExternalId) {
```
kbottla, line 32: Is this method required? I see it is only used in tests

src/main/java/uk/gov/pay/webhooks/message/dao/entity/WebhookMessageEntity.java
```diff
@@ -30,27 +30,27 @@
 
 @NamedQuery(
         name = WebhookMessageEntity.MESSAGE_BY_WEBHOOK_ID_AND_MESSAGE_ID,
-        query = "select m from WebhookMessageEntity m where webhookEntity.externalId = :webhookId and externalId = :messageId"
+        query = "select m from WebhookMessageEntity m where webhookEntity = :webhook and externalId = :messageId"
```
kbottla, line 33: Is it possible to use condition `webhookEntity.id.= :id` for all queries so it is clear about the field being queried? we are passing entire Webhooks entity to these named queries which may not be needed.

sfount: I'm not sure if JPA will assume it needs to cross join to get the entity in order to compare an attribute of it, I could try. 

Using this method I don't need to add an accessor for the internal identifier on webhooks and just let JPA figure out its the `@ID` field that makes it unique.

I have checked the output queries just to be sure we're OK from a performance perspective and it does behave as expected (without a cross join, webhook entity marked as `FetchType.LAZY`)

Example from debugger
```
/* WebhookMessage.count_messages_by_webhook_id */ 
select count(webhookmes0_.id) as col_0_0_ from webhook_messages webhookmes0_ where webhookmes0_.webhook_id=$1
```

src/test/java/uk/gov/pay/webhooks/message/dao/WebhookMessageDaoTest.java
sfount: Yeah good question, this is a test using dropwizard extension helpers that emulate database behaviour but it doesn't actually spin up either a database or the app itself (rather constructs the DAO and passes in the emulated jdbi). I'm not sure exactly where the line would be drawn but this is orders of magnitude faster than spinning up real instances using test containers, for example

kbottla (approved):

PR #721, opened 2023-03-23
PP-10834 Remove total from webhook messages pagination

sfount:
The admin tool should no longer provide a "Last" link when showing webhook messags pagination. The last link is calculated by taking the total number of webhook messages and dividing by the current page size.

Remove the total calculation when webhook messages are listed, this should avoid any costly lookups if there is any significant scale on webhook messages in the given webhook message TTL.

Depends on https://github.com/alphagov/pay-selfservice/pull/3849


SandorArpa (approved):
LOKTM

PR #776, opened 2023-04-27
PP-10835: Endpoint for expiring webhook messages and webhook delivery attempts

oswaldquek:


src/main/java/uk/gov/pay/webhooks/webhook/WebhookService.java
```diff
@@ -157,19 +158,32 @@ private boolean webhookHasSubscriptionForEvent(WebhookEntity webhook, InternalEv
                 .anyMatch(internalEventNames -> internalEventNames.contains(event.eventType()));
     }
 
-    public void deleteWebhookMessagesOlderThan() {
+    public void expireWebhookMessages() {
```
kbottla, line 161: I think `delete` in the name is fine. `expire` could be confused with status being expired

src/main/java/uk/gov/pay/webhooks/webhook/WebhookService.java
```diff
@@ -157,19 +158,32 @@ private boolean webhookHasSubscriptionForEvent(WebhookEntity webhook, InternalEv
...
         int maxAgeOfMessages = webhookMessageDeletionConfig.getMaxAgeOfMessages();
         int maxNumOfMessagesToExpire = webhookMessageDeletionConfig.getMaxNumOfMessagesToExpire();
+
+        List<WebhookMessageEntity> candidateWebhookMessagesForDeletion = webhookMessageDao.getWebhookMessagesOlderThan(maxAgeOfMessages);
```
kbottla, line 165: Any reason we are querying for a list of IDs and deleting webhook messages?

Can we not delete webhook_messages and cascade by date?

src/main/java/uk/gov/pay/webhooks/webhook/WebhookService.java
```diff
@@ -157,19 +158,32 @@ private boolean webhookHasSubscriptionForEvent(WebhookEntity webhook, InternalEv
...
+        List<WebhookDeliveryQueueEntity> candidateDeliveryQueueEntitiesForDeletion = 
+                webhookDeliveryQueueDao.getWebhookDeliveryQueueEntitiesOlderThan(maxAgeOfMessages);
+        List<Long> webhookMessageEntityIdsReferencedByCandidateDeliveryQueueEntities = 
+                candidateDeliveryQueueEntitiesForDeletion.stream().map(entity -> entity.getWebhookMessageEntity().getId()).collect(Collectors.toList());
         
-        var webhookMessages = webhookMessageDao.getWebhookMessagesOlderThan(maxAgeOfMessages);
-        int webhookMessagesDeleted = webhookMessageDao.deleteMessages(
-                webhookMessages.stream().limit(maxNumOfMessagesToExpire));
-        LOGGER.info(format("%s webhook messages were deleted.", webhookMessagesDeleted));
```
kbottla, line 167: This might be useful to know

src/main/java/uk/gov/pay/webhooks/webhook/WebhookService.java
```diff
@@ -157,19 +158,32 @@ private boolean webhookHasSubscriptionForEvent(WebhookEntity webhook, InternalEv
...
-
-        var deliveryQueueEntities = webhookDeliveryQueueDao.getWebhookDeliveryQueueEntitiesOlderThan(maxAgeOfMessages);
-        int deliveryQueueEntitiesDeleted = webhookDeliveryQueueDao.deleteDeliveryQueueEntries(
-                deliveryQueueEntities.stream().limit(maxNumOfMessagesToExpire));
+        var webhookDeliveryQueueEntitiesToDelete = candidateDeliveryQueueEntitiesForDeletion.stream().limit(maxNumOfMessagesToExpire).toList();
```
kbottla, line 171: Don't think we need to limit the number of messages from the `queue` table.

We can remove entries from `webhooks_message` that are older than 7 days and by `maxNumOfMessagesToExpire`. We can remove all associated `queue` entries.

src/main/resources/config/config.yaml
```diff
@@ -80,6 +80,10 @@ queueMessageReceiverConfig:
   numberOfThreads: ${QUEUE_MESSAGE_RECEIVER_NUMBER_OF_THREADS:-3}
   messageRetryDelayInSeconds: ${QUEUE_MESSAGE_RETRY_DELAY_IN_SECONDS:-900}
 
+webhookMessageDeletionConfig:
+  maxAgeOfMessages: ${MAXIMUM_AGE_OF_MESSAGE:-7}
+  maxNumOfMessagesToExpire: ${MAXIMUM_NUMBER_OF_MESSAGES_TO_EXPIRE_PER_JOB:-7500}
```
kbottla, line 85: The default config can be 15000 which will be a number of entries from the `webhook_messages` table. We can remove all associated entries from the queue table.

src/main/java/uk/gov/pay/webhooks/webhook/WebhookService.java
oswaldquek: 🤦 Delete cascade would be much easier. Might even warrant an entirely new PR.

src/main/java/uk/gov/pay/webhooks/webhook/WebhookService.java
```diff
@@ -157,19 +156,12 @@ private boolean webhookHasSubscriptionForEvent(WebhookEntity webhook, InternalEv
...
-        int webhookMessagesDeleted = webhookMessageDao.deleteMessages(
-                webhookMessages.stream().limit(maxNumOfMessagesToExpire));
-        LOGGER.info(format("%s webhook messages were deleted.", webhookMessagesDeleted));
-
-        var deliveryQueueEntities = webhookDeliveryQueueDao.getWebhookDeliveryQueueEntitiesOlderThan(maxAgeOfMessages);
-        int deliveryQueueEntitiesDeleted = webhookDeliveryQueueDao.deleteDeliveryQueueEntries(
-                deliveryQueueEntities.stream().limit(maxNumOfMessagesToExpire));
-        LOGGER.info(format("%s webhook delivery queue entities were deleted.", deliveryQueueEntitiesDeleted));
 
+        List<WebhookMessageEntity> webhookMessagesToDelete = webhookMessageDao.getWebhookMessagesOlderThan(maxAgeOfMessages);
```
kbottla, line 163: Any reason webhooks messages need to be retrieved and deleted? Can we not delete without querying?

src/test/java/uk/gov/pay/webhooks/webhook/resource/WebhookResourceIT.java
```diff
@@ -206,6 +216,111 @@ public void notFoundShouldReturn404() {
...
+                        .then()
+                        .statusCode(Response.Status.NOT_FOUND.getStatusCode()));
+    }
+
+    private List<String> setupThreeWebhookMessagesThatShouldNotBeDeleted() {
+        DateFormat df = new SimpleDateFormat("yyyy-MM-dd");
+        String date = df.format(Date.from(OffsetDateTime.now().minusDays(1).toInstant()));
+        List<String> webhookMessageExternalIds = List.of("thirteenth-message-external-id", "fourteenth-message-external-id", "fifteenth-message-external-id");
+        app.getJdbi().withHandle(h -> h.execute("""
+                            INSERT INTO webhook_messages VALUES
```
kbottla, line 262: Can this SQL be in DbTestHelper?

oswaldquek: I can't see any existing DbTestHelper class so do you mean a new class?
I'm following the [existing pattern](https://github.com/alphagov/pay-webhooks/blob/main/src/test/java/uk/gov/pay/webhooks/webhook/resource/WebhookResourceIT.java#L221-L260) here.

kbottla: It is named `DatabaseTestHelper`.

src/main/java/uk/gov/pay/webhooks/webhook/WebhookService.java
oswaldquek: You're right, messages don't need to be retrieved.

oswaldquek: > Any reason webhooks messages need to be retrieved and deleted? Can we not delete without querying?

Actually this won't work because we have a `MAXIMUM_NUMBER_OF_MESSAGES_TO_EXPIRE_PER_JOB` env var and we can't use a `DELETE` in conjunction with `LIMIT` in postgresql.

kbottla: Could use something like `delete from webhook_messages where id in (select id from webhook_messages where created_date < '2022-12-31' limit 15000);`

src/test/java/uk/gov/pay/webhooks/webhook/resource/WebhookResourceIT.java
```diff
@@ -206,6 +216,111 @@ public void notFoundShouldReturn404() {
...
+    @Test
+    public void shouldDeleteSomeWebhookMessages() {
+        var webhookExternalId = "a-webhook-external-id";
+        WebhookMessageExternalIds webhookMessageExternalIds = setupWebhookWithMessagesExpectedToBePartiallyDeleted(webhookExternalId);
+        List<String> expectedWebhookExternalIdsNotDeleted = webhookMessageExternalIds.notDeleted;
+        List<String> expectedWebhookMessageExternalIds = setupThreeWebhookMessagesThatShouldNotBeDeleted(); // maxAgeOfMessages=7 so these webhook messages should not be deleted
+
+        given().port(port)
+                .contentType(JSON)
+                .post("/v1/webhook/tasks/delete_messages")
```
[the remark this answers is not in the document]
oswaldquek: I had talked about this with @sfount before and we agreed it could go in here.

src/main/java/uk/gov/pay/webhooks/message/dao/WebhookMessageDao.java
```diff
@@ -63,16 +59,11 @@ private int calculateFirstResult(int page) {
         return (page - 1) * WEBHOOK_MESSAGES_PAGE_SIZE;
     }
 
-    public List<WebhookMessageEntity> getWebhookMessagesOlderThan(int days) {
-        Preconditions.checkArgument(days > 0, "Can only get webhook messages older than 0 days.");
-        return namedTypedQuery(WebhookMessageEntity.MESSAGES_OLDER_THAN_X_DAYS)
+    public int deleteMessages(int days, int maxNumOfMessagesToDelete) {
+        return currentSession().createNativeQuery("delete from webhook_messages where id in " +
+                        "(select id from webhook_messages where created_date < :datetime limit :maxNumOfMessagesToDelete)")
```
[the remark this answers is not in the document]
oswaldquek: I don't think it matters as we're already getting the records before a certain date.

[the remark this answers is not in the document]
oswaldquek: We'll be running this query every thirty minutes too so I don't think this matters. We can certainly revisit if we run into performance problems.

src/test/java/uk/gov/pay/webhooks/webhook/resource/WebhookResourceIT.java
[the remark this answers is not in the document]
oswaldquek: The `/v1/webhook` comes from the fact that we've decided (and agreed with @sfount) that the endpoint can go into WebhookResource.java which prepends all endpoints with `/v1/webhook`. I don't think we need to make a big deal about this.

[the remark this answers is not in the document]
oswaldquek: We especially don't need to make a big deal about this because it's an internal endpoint which can configure in pay-infra and not customer facing.

sfount (dismissed):
Quickly looking at `src/main/resources/migrations/0013_alter_table_webhook_delivery_queue_add_on_delete_cascade.sql` is that something we want to move out in case we run into any issues on the migration? Because this endpoint won't be invoked right away I can see arguments both ways but it might make it simpler for now.

I've had a quick chat with @kbottla and @oswaldquek and it looks like there are a few patterns we can clean up now we've got multiple examples of them across this app/ other apps, I've made stories to address these in particular
- extracting test helpers to redude boilerplates across tests
- looking into the format for url endpoints on endpoints called by administrative tasks

sfount (approved):

sfount:
Migration moved out to https://github.com/alphagov/pay-webhooks/pull/784

PR #781, opened 2023-05-02
PP-10836 Add index on resource_external_id

stephencdaly:
Add a btree index to the webhook_messages table on the resource_external_id column, so it is possible to search for messages by the resource (payment/refund etc) id.


SandorArpa (approved):
LGTM

PR #784, opened 2023-05-03
PP-10835: Add "ON DELETE CASCADE" to webhook_messages

oswaldquek:
In order to delete from webhook_messages, corresponding entries in webhook_delivery_queue need to be deleted. It would be more expedient to delete from webhook_delivery_queue by setting "ON DELETE CASCADE" to webhook_messages rather than deleting from both tables separately.


sfount (approved):
Because there's an index on this column I think the constraint should be able to be updated without locking the table for too long, I'm not 100% though. If we run into issues we might need to concurrently add an index and reference it from this, lets keep an eye on this in test before progressing.

PR #787, opened 2023-05-04
PP-10836 Allow filtering messages by resource ID

stephencdaly:
- Accept query parameter "resource_id" to filter returned messages by a payment/refund etc.
- Move query parameters to a POJO using @BeanParam annotation
- Use dynamically constructed hibernate query rather than named query
- Remove `count` method from DAO which is unused since changes to pagination to remove the "total" from the API response.


SandorArpa (approved):
LOKTM

PR #797, opened 2023-05-11
PP-10538 Add logging for receiving events from SNS

stephencdaly:
Add logging when a message is successfully processed from the SQS queue that subscribes to SNS topics that events are published to from ledger.


SandorArpa (approved):
LOKTM

PR #813, opened 2023-05-18
PP-11019 Serealise gateway account id on webhooks

sfount:
Depends on https://github.com/alphagov/pay-webhooks/pull/812

Add gateway account id to the webhooks entity to reflect the column
added to the database table.

Include on the resource model and require when creating a new webhook.

This field will be used to lookup webhooks and webhook data (message,
attempts, etc.) as a separate PR to allow the data to be appropriately
backfilled in environments before being relied on.


alexbishop1 (approved):
💯

PR #816, opened 2023-05-19
PP-10940 Set a connection manager for webhook message sender client

sfount:
Use a thread safe pooling connection manager the http client used to send webhook messages. This should be configured to allow at least the number of webhook message sending threads number of connections to lease.


src/main/java/uk/gov/pay/webhooks/app/WebhookMessageSendingQueueProcessorConfig.java
```diff
@@ -29,14 +33,18 @@ public int getThreadDelayInMilliseconds() {
...
+    public int getWebhookMessageSendingQueueProcessorNumberOfThreads() {
+        return webhookMessageSendingQueueProcessorNumberOfThreads;
     }
 
     public int getInitialDelayInMilliseconds() {
         return initialDelayInMilliseconds;
     }
 
+    public int getConnectionPoolNumberOfThreads() {
+        return connectionPoolNumberOfThreads;
```
sfount, line 45: I think calling these "threads" though isn't quite right, it's essentially the maximum number of simultaneous http connections that can be available in the pool at any given time for a client to use, the client can access those connections from any thread.

I think something like "httpClientConnectionPoolSize" or something like that might better reflect it

sfount (commented):
Implementation looks great, question about the naming

src/main/java/uk/gov/pay/webhooks/app/WebhookMessageSendingQueueProcessorConfig.java
stephencdaly: Updated

sfount (commented):
✅

stephencdaly (approved):

PR #817, opened 2023-05-19
PP-10940 Emit webhook messages in reverse order

sfount:
During performance testing, we observed that not ordering the delivery queue containing webhook messages which need to be sent leads to a recency bias, under particularly high volume messages can be ignored until the backlog has been cleard and volume reduced. This could theoretically lead to messages never being sent in the worst case.

Order the `send_at` property by the older row that meets the criteria to be sent.

In order to the performance cost described in
https://github.com/alphagov/pay-webhooks/pull/574, add a composite index to allow postres to ignore any entries other than 'PENDING' and efficiently order (and limit) by time.

There is also now an X day retention period on messages (and their delivery queue entries) which should help keep the table timely.

A backlog, under load, with no ordering of messages to send. Note the 3 to 4 minute time delay on messages finally being sent after the load subsides.
Screenshot 2023-05-19 at 16 40 28

A backlog, under load, with ordering of messages. Note a more consistent time to address the backlog once, the delay is spread among messages with a maximum of 1.5 minutes.
Screenshot 2023-05-19 at 16 41 01

### DIFF
--- a/src/main/java/uk/gov/pay/webhooks/deliveryqueue/dao/WebhookDeliveryQueueEntity.java
+++ b/src/main/java/uk/gov/pay/webhooks/deliveryqueue/dao/WebhookDeliveryQueueEntity.java
@@ -30,7 +30,7 @@ import java.util.Optional;
 
 @NamedQuery(
         name = WebhookDeliveryQueueEntity.NEXT_TO_SEND,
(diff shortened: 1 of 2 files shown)


stephencdaly (approved):

james-peacock-gds (approved):
LGTM - I saw the code change applied and pushed from a branch alongside the indexing change resulting in the effect described.

PR #820, opened 2023-05-23
PP-11027 Lookup relevant webhook subscriptions by gateway account ID

sfount:
Gateway account IDs were added in https://github.com/alphagov/pay-webhooks/pull/812 and https://github.com/alphagov/pay-webhooks/pull/813.

They should also [be available on the payment events schema](https://github.com/alphagov/pay-connector/pull/4525).

Until the data model across Pay shifts to being service-centric, use the gateway account ID to determine if there are any webhooks subscribed for a given event (this will allow multiple test card payment accounts to be associated with a single service without causing any issues in the webhooks mechanism or reporting on webhooks data, for the time being).


stephencdaly (approved):
LGTM!

PR #835, opened 2023-06-01
PP-11059 Stale Connection Monitor

james-peacock-gds:
Context: Performance testing of webhooks identified client attempts to re-use expired connections.
- add stale connection monitor thread
- add idle connection time-to-live to config


src/main/java/uk/gov/pay/webhooks/app/IdleConnectionMonitorThread.java
```diff
@@ -0,0 +1,39 @@
...
+
+import org.apache.http.conn.HttpClientConnectionManager;
+import org.apache.http.impl.conn.PoolingHttpClientConnectionManager;
+
+import java.util.concurrent.TimeUnit;
+
+public class IdleConnectionMonitorThread extends Thread {
+    private final HttpClientConnectionManager connMgr;
+    private volatile boolean shutdown;
+    private final long idleConnectionTimeToLive;
```
alexbishop1, line 11: This could be a `Duration` (rather than a `long`) and just be converted every time it’s run.

src/main/java/uk/gov/pay/webhooks/app/WebhooksModule.java
```diff
@@ -85,14 +82,24 @@ public CloseableHttpClient httpClient() {
...
                 null,
                 SSLConnectionSocketFactory.getDefaultHostnameVerifier()
         );
-
+        
+        IdleConnectionMonitorThread staleMonitor
+                = new IdleConnectionMonitorThread(poolingConnManager, configuration.getWebhookMessageSendingQueueProcessorConfig().getConnectionPoolIdleConnectionTimeToLive().toSeconds());
+        staleMonitor.start();
+        try {
+            staleMonitor.join(1000);
```
alexbishop1, line 97: Can we un-magic this number?

src/main/java/uk/gov/pay/webhooks/app/WebhooksModule.java
```diff
@@ -85,14 +82,24 @@ public CloseableHttpClient httpClient() {
...
+        }
+        catch (InterruptedException ex) {
+            ex.printStackTrace();
```
alexbishop1, line 100: I think we want to this to be logged somewhere.

alexbishop1 (changes requested):

src/main/java/uk/gov/pay/webhooks/app/WebhooksModule.java
kbottla, line 100: Is there another way to log this rather than using printStackTrace?

src/main/java/uk/gov/pay/webhooks/app/IdleConnectionMonitorThread.java
```diff
@@ -0,0 +1,39 @@
...
+        super();
+        this.connMgr = connMgr;
+        this.idleConnectionTimeToLive = idleConnectionTimeToLive;
+    }
+    @Override
+    public void run() {
+        try {
+            while (!shutdown) {
+                synchronized (this) {
+                    wait(1000);
```
kbottla, line 24: Not a blocker but is it possible to have wait time as config variable?

src/main/java/uk/gov/pay/webhooks/app/WebhooksModule.java
kbottla, line 97: I don't understand why a join is required. Can the commit message be amended to explain how this is expected to work or add a comment if appropriate?

PR #854, opened 2023-06-13
PP-11059 Idle Connection Monitor

james-peacock-gds:
Context: Performance testing of webhooks identified client attempts to re-use expired connections to send webhook messages
- add IdleConnectionMonitor to identify and close idle/expired connections
- add configuration options for IdleConnectionMonitor


src/main/resources/config/config.yaml
```diff
@@ -93,6 +93,10 @@ webhookMessageSendingQueueProcessorConfig:
   threadDelayInMilliseconds: ${WEBHOOK_MESSAGE_SENDING_QUEUE_THREAD_DELAY_IN_MILLISECONDS:-1000}
   httpClientConnectionPoolSize: ${WEBHOOK_MESSAGE_SENDER_CONNECTION_POOL_SIZE:-20}
   connectionPoolTimeToLive: ${WEBHOOK_MESSAGE_SENDER_CONNECTION_POOL_TIME_TO_LIVE:-60s}
+  connectionPoolIdleConnectionTimeToLive: ${WEBHOOK_MESSAGE_SENDER_CONNECTION_POOL_IDLE_CONNECTION_TIME_TO_LIVE:-5s}
   requestTimeout: ${WEBHOOK_MESSAGE_SENDER_REQUEST_TIMEOUT:-3s}
+  idleConnectionMonitorEnabled: ${WEBHOOK_MESSAGE_SENDER_IDLE_CONNECTION_MONITOR_ENABLED:-true}
```
kbottla, line 98: Would be good to update test-config.yaml too so the connection manager can be enabled in tests and see if there are any issues for it starting with the test config.

kbottla, line 98: Do we want to enable it straightway in production as the default is `true`?

james-peacock-gds: done - also added the other required configuration options

james-peacock-gds: good point, done.

kbottla (approved):

PR #921, opened 2023-08-03
PP-11223: Set up prometheus metrics

oswaldquek:
Set up a prometheus metric endpoint /metrics.
Append ECS metadata info (via the PrometheusDefaultLabelSampleBuilder class in pay-java-commons) as default labels.


pom.xml
```diff
@@ -90,6 +91,21 @@
 
     <dependencies>
         <!-- Main dependencies that use a BOM so no explicit versions needed -->
+        <dependency>
```
nlsteers, line 94: given the above comment, these new deps should be moved down to where deps are declared with specific versions.

oswaldquek: Fair point - fixed.

PR #958, opened 2023-09-01
PP-11406: Log domain even if request timed out

oswaldquek:
If the request timed out, the request was still attempted. If the request was attempted, the domain is logged in the egress logs. Currently, if the reqest is timed out, the domain isn't logged in the application logs, which leads to a mismatch between the domains logged in the webhook application logs and egress logs. This causes the "Compare webhooks and webhooks-egress target domains" alert to fire unnecessarily. This commit logs the domain even if the request times out, and uses `URI.create` which does not throw a checked exception which makes the code a bit nicer.

### DIFF
--- a/src/main/java/uk/gov/pay/webhooks/deliveryqueue/managed/SendAttempter.java
+++ b/src/main/java/uk/gov/pay/webhooks/deliveryqueue/managed/SendAttempter.java
@@ -19,7 +19,7 @@ import javax.inject.Inject;
 import javax.ws.rs.core.Response;
 import java.io.IOException;
 import java.net.SocketTimeoutException;
-import java.net.URL;
+import java.net.URI;
 import java.net.http.HttpTimeoutException;
 import java.security.InvalidKeyException;
 import java.time.Duration;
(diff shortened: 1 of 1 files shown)


src/main/java/uk/gov/pay/webhooks/deliveryqueue/managed/SendAttempter.java
```diff
@@ -107,7 +108,6 @@ public void attemptSend(WebhookDeliveryQueueEntity queueItem) {
             );
             handleResponse(queueItem, DeliveryStatus.WILL_NOT_SEND, null, "Violates security rules", retryCount, start);
         } catch (Exception e) {
-            var responseTime = Duration.between(start, instantSource.instant());
```
katstevens, line 110: Ah, should this have been removed during a previous refactor at some point?

oswaldquek: Probably. I thought I might as well remove it now as the variable is not being used.

katstevens (approved):

PR #964, opened 2023-09-06
PP-11228: Remove graphite metrics sending and config

jfharden:
## WHAT YOU DID
1. Remove graphite metrics configuration and sending
2. Rename GRAPHITE_SENDING_PERIOD_SECONDS to METRICS_COLLECTION_PERIOD_SECONDS, and set it to the appropriate value.

## How to test

Check build passes and explicitly wait for the e2e tests to pass

### Logging

- [X] only emit log lines at ERROR level which require immediate attention from a support engineer. These will trigger a zendesk alert.

### DIFF
--- a/src/main/java/uk/gov/pay/webhooks/app/WebhooksApp.java
+++ b/src/main/java/uk/gov/pay/webhooks/app/WebhooksApp.java
@@ -51,7 +51,7 @@ public class WebhooksApp extends Application<WebhooksConfig> {
     }
     
     private static final String SERVICE_METRICS_NODE = "webhooks";
-    private static final int METRICS_COLLECTION_PERIOD_SECONDS = 10;
+    private static final int METRICS_COLLECTION_PERIOD_SECONDS = 30;
 
     private final HibernateBundle<WebhooksConfig> hibernate = new HibernateBundle<>(
             WebhookEntity.class,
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/pay/webhooks/app/WebhooksApp.java
```diff
@@ -54,7 +51,7 @@ public static void main(String[] args) throws Exception {
     }
     
     private static final String SERVICE_METRICS_NODE = "webhooks";
-    private static final int GRAPHITE_SENDING_PERIOD_SECONDS = 10;
+    private static final int METRICS_COLLECTION_PERIOD_SECONDS = 10;
```
kbottla, line 54: Should this be 30 similar to other app changes? https://github.com/alphagov/pay-products/pull/1210/files#diff-4f6066de42bd7de905cdc3cdfa57b9396cfa960c9ebea66f2559f7c21e74f6caR55

jfharden: Yeah it should! Thanks

jfharden: Updated 👍

PR #1026, opened 2023-10-20
PP-11652 Upgrade from WireMock 2.35.1 to 3.2.0

alexbishop1:
with @SandorArpa


SandorArpa (approved):
I’m confident that this change does not require further review and meets the secure coding requirements of GOV.UK Pay

PR #1027, opened 2023-10-20
PP-11652 Fix logging to Sentry

alexbishop1:
with @SandorArpa


SandorArpa (approved):
I’m confident that this change does not require further review and meets the secure coding requirements of GOV.UK Pay

PR #1082, opened 2023-11-23
PP 10972 refactor adhoc test sql database test helper

olatomgds:
### Description

There are a number of tests that write ad-hoc SQL (executed by JDBI taken from the running DropWizard instance) for setting up the environment. In a number of these tests the setup and entities created are shared. 

Identify any places this setup could be extracted into the DatabaseTestHelper (see https://github.com/alphagov/pay-webhooks/blob/main/src/test/java/uk/gov/pay/webhooks/util/DatabaseTestHelper.java#L5) and shared between different tests. This should allow future tests to make use of the same boilerplate code and focus on the logic under test.

The change also included the following:
- De-duplicating SQL statements and updating the code references
- Grouping methods referencing same database table together for ease of future rework . code maintenance
- Combining multiple database calls in a method into one call for performance of test execution
- Renaming method names to reflect the CRUD operation and associated table
- Primarily abstracting methods out of test classes to the DB Test Helper
- Rework of redundant test data or parameters logic
- Rename of misleading method names

### How to test
Review of code and comparison of previous version with current and running all impacted tests.

### DIFF
--- a/src/test/java/uk/gov/pay/webhooks/util/DatabaseTestHelper.java
+++ b/src/test/java/uk/gov/pay/webhooks/util/DatabaseTestHelper.java
@@ -13,6 +13,10 @@ public class DatabaseTestHelper {
     public static DatabaseTestHelper aDatabaseTestHelper(Jdbi jdbi) {
         return new DatabaseTestHelper(jdbi);
     }
+
+    public void executeSql(String sql) {
+        jdbi.withHandle(h -> h.execute(sql));
+    }
     
     public void addWebhookWithSubscription(String webhookExternalId, String serviceExternalId, String callbackUrl, String gatewayAccountId) {
         jdbi.withHandle(h -> h.execute("INSERT INTO webhooks VALUES (1, '2022-01-01', '%s', 'signing-key', '%s', false, '%s', 'description', 'ACTIVE', '%s')".formatted(webhookExternalId, serviceExternalId, callbackUrl, gatewayAccountId)));
@@ -20,6 +24,18 @@ public class DatabaseTestHelper {
         jdbi.withHandle(h -> h.execute("INSERT INTO webhook_subscriptions VALUES (1, (SELECT id FROM event_types WHERE name = 'card_payment_refunded'))"));
     }
 
+    public void addWebhookMessageLastDeliveryStatusIsConsistent(String serviceExternalId, String gatewayAccountId, int wireMockPort) {
+        jdbi.withHandle(h -> h.execute("INSERT INTO webhooks VALUES (1, '2022-01-01', 'webhook-external-id-succeeds', 'signing-key', '%s', false, 'http://localhost:%d/a-working-endpoint', 'description', 'ACTIVE', '%s')".formatted(serviceExternalId, wireMockPort, gatewayAccountId)));
+        jdbi.withHandle(h -> h.execute("INSERT INTO webhooks VALUES (2, '2022-01-01', 'webhook-external-id-fails', 'signing-key', '%s', false, 'http://localhost:%d/a-failing-endpoint', 'description', 'ACTIVE', '%s')".formatted(serviceExternalId, wireMockPort, gatewayAccountId)));
+        jdbi.withHandle(h -> h.execute("INSERT INTO webhook_subscriptions VALUES (1, (SELECT id FROM event_types WHERE name = 'card_payment_succeeded'))"));
+        jdbi.withHandle(h -> h.execute("INSERT INTO webhook_subscriptions VALUES (2, (SELECT id FROM event_types WHERE name = 'card_payment_succeeded'))"));
+    }
+
+    public void addDeliveryStatusEnumIsConsistentWithDatabase() {
+        jdbi.withHandle(h -> h.execute("INSERT INTO webhooks VALUES (1, '2022-01-01', 'webhook-external-id', 'signing-key', 'service-id', true, 'https://callback-url.test', 'description', 'ACTIVE')"));
+        jdbi.withHandle(h -> h.execute("INSERT INTO webhook_messages VALUES (1, 'message-external-id', '2022-01-01', 1, '2022-01-01', 1, '{}', 'transaction-external-id', 'payment')"));
+    }
+    
     public void truncateAllData() {
         jdbi.withHandle(h -> h.createScript(
                 "TRUNCATE TABLE webhooks CASCADE; "
--- a/src/test/java/uk/gov/pay/webhooks/util/DatabaseTestHelper.java
+++ b/src/test/java/uk/gov/pay/webhooks/util/DatabaseTestHelper.java
@@ -6,9 +6,6 @@ import uk.gov.pay.webhooks.deliveryqueue.DeliveryStatus;
 import java.util.List;
 
 public class DatabaseTestHelper {
-
-    public static final String EVENT_TYPE_SELECT_QUERY_CARD_PAYMENT_SUCCEEDED = "SELECT id FROM event_types WHERE name = 'card_payment_succeeded'";
-    public static final String EVENT_TYPE_SELECT_QUERY_CARD_PAYMENT_REFUNDED = "SELECT id FROM event_types WHERE name = 'card_payment_refunded'";
     private Jdbi jdbi;
 
     private DatabaseTestHelper(Jdbi jdbi) {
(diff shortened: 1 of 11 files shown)


src/test/java/uk/gov/pay/webhooks/util/DatabaseTestHelper.java
```diff
@@ -13,17 +19,163 @@ private DatabaseTestHelper(Jdbi jdbi) {
...
+    public void addWebhookMessageLastDeliveryStatusIsConsistent(String serviceExternalId, String gatewayAccountId, int wireMockPort) {
+        jdbi.withHandle(h -> h.execute("""
+                    INSERT INTO webhooks VALUES
+                    (1, '2022-01-01', 'webhook-external-id-succeeds', 'signing-key', '%s', false, 'http://localhost:%d/a-working-endpoint', 'description', 'ACTIVE', '%s'),
+                    (2, '2022-01-01', 'webhook-external-id-fails', 'signing-key', '%s', false, 'http://localhost:%d/a-failing-endpoint', 'description', 'ACTIVE', '%s')
+                """.formatted(
+                serviceExternalId, wireMockPort, gatewayAccountId,
+                serviceExternalId, wireMockPort, gatewayAccountId)
+        ));
+    }
```
alexbishop1, line 36: This is a bit specific. It looks like this method is inserting two webhooks with different webhook external IDs. So we could have…

```java
public void addWebhook(String serviceExternalId, String gatewayAccountId, String webhookExternalId, int wireMockPort) {
    jdbi.withHandle(h -> h.execute("""
            INSERT INTO webhooks VALUES
            (1, '2022-01-01', '%s', 'signing-key', '%s', false, 'http://localhost:%d/a-working-endpoint', 'description', 'ACTIVE', '%s'))
            """.formatted(webhookExternalId, serviceExternalId, wireMockPort, gatewayAccountId);
}
```

… and just call it twice from the test with different arguments.

src/test/java/uk/gov/pay/webhooks/util/DatabaseTestHelper.java
```diff
@@ -13,17 +19,163 @@ private DatabaseTestHelper(Jdbi jdbi) {
...
+    
+    public void addWebhookSubscription() {
+        jdbi.withHandle(h -> h.execute("""
+                    INSERT INTO webhook_subscriptions VALUES
+                    (1, (SELECT id FROM event_types WHERE name = 'card_payment_succeeded')),
+                    (1, (SELECT id FROM event_types WHERE name = 'card_payment_refunded'))
+                """
+        ));
+    }
```
alexbishop1, line 52: Similar to the above, we could make this method `addWebhookSubscription(String eventName)` and call it twice from the test.

src/test/java/uk/gov/pay/webhooks/util/DatabaseTestHelper.java
```diff
@@ -13,17 +19,163 @@ private DatabaseTestHelper(Jdbi jdbi) {
...
+
+    public void addWebhookSubscriptionsMessageLastDeliveryStatusIsConsistent() {
+        jdbi.withHandle(h -> h.execute("""
+                    INSERT INTO webhook_subscriptions VALUES
+                    (1, (SELECT id FROM event_types WHERE name = 'card_payment_succeeded')),
+                    (2, (SELECT id FROM event_types WHERE name = 'card_payment_succeeded'))
+                """
+        ));
+    }
```
alexbishop1, line 61: I think this method would be unnecessary if the previous one is parameterised as suggested.

src/test/java/uk/gov/pay/webhooks/util/DatabaseTestHelper.java
```diff
@@ -13,17 +19,163 @@ private DatabaseTestHelper(Jdbi jdbi) {
...
+                    (6, 'sixth-message-external-id', '2022-01-01', 1, '2022-01-01', 1, '{}', null, null, null),
+                    (7, 'seventh-message-external-id', '2022-01-01', 1, '2022-01-01', 1, '{}', null, null, null),
+                    (8, 'eighth-message-external-id', '2022-01-01', 1, '2022-01-01', 1, '{}', null, null, null),
+                    (9, 'ninth-message-external-id', '2022-01-01', 1, '2022-01-01', 1, '{}', null, null, null),
+                    (10, 'tenth-message-external-id', '2022-01-01', 1, '2022-01-01', 1, '{}', null, null, null),
+                    (11, 'eleventh-message-external-id', '2022-01-01', 1, '2022-01-01', 1, '{}', null, null, null),
+                    (12, 'twelfth-message-external-id', '2022-01-01', 1, '2022-01-01', 1, '{}', null, null, null)
+                """.formatted(messageExternalId)
+        ));
+    }
```
alexbishop1, line 119: I think all of these could be replaced by a small number of parameterised methods that are called multiple times from the test as necessary.

src/test/java/uk/gov/pay/webhooks/util/DatabaseTestHelper.java
```diff
@@ -13,17 +19,163 @@ private DatabaseTestHelper(Jdbi jdbi) {
...
+                    (8, '2022-01-01', '2022-01-01', '404', 404, 6, 'PENDING', null),
+                    (9, '2022-01-01', '2022-01-01', '404', 404, 7, 'PENDING', null),
+                    (10, '2022-01-01', '2022-01-01', '404', 404, 8, 'PENDING', null),
+                    (11, '2022-01-01', '2022-01-01', '404', 404, 9, 'PENDING', null),
+                    (12, '2022-01-01', '2022-01-01', '404', 404, 10, 'PENDING', null),
+                    (13, '2022-01-01', '2022-01-01', '404', 404, 11, 'PENDING', null),
+                    (14, '2022-01-01', '2022-01-01', '404', 404, 12, 'PENDING', null)
+                """
+        ));
+    }
```
alexbishop1, line 174: You know what I’m going to say here! 😛

alexbishop1 (changes requested):
I love the use of text blocks!

I think some of the messages could be parameterised and called multiple times from the tests to make them more reusable.

olatomgds:
Thanks for the feedback, they have been all addressed and additional enhancement done.

alexbishop1 (changes requested):
This looks a lot clearer!

As stated in the comments, we don’t use Lombok in Pay projects but I think the builder pattern does work well here. Could we manually (or IDE-generate) the builder methods rather than using Lombok?

alexbishop1 (changes requested):
Looks good. I think we could make a couple of the classes into records, which would make them more concise, so please do that if possible.

src/test/java/uk/gov/pay/webhooks/util/DatabaseTestHelper.java
```diff
@@ -13,17 +20,271 @@ private DatabaseTestHelper(Jdbi jdbi) {
...
+        public String getLive() {
+            return live;
+        }
+
+        public String getGatewayAccountId() {
+            return gatewayAccountId;
+        }
+    }
+
+    public static class WebhookSubscription implements Serializable {
```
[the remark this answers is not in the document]
olatomgds: Yes, found a way to add mutable methods to the java record, choice of record looks better. Thanks.

src/test/java/uk/gov/pay/webhooks/util/DatabaseTestHelper.java
```diff
@@ -13,17 +19,101 @@ private DatabaseTestHelper(Jdbi jdbi) {
...
+            String externalId,
+            String createdDate,
+            int webhookId,
+            String eventDate,
+            int eventType,
+            String resource,
+            String resourceExternalId,
+            String resourceType,
+            DeliveryStatus deliveryStatus) {
+        public WebhookMessage setWebhookMessageAndExternalId(int webhookMessageId, String externalId) {
```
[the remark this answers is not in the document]
olatomgds: Grand

src/test/java/uk/gov/pay/webhooks/util/DatabaseTestHelper.java
```diff
@@ -13,17 +19,101 @@ private DatabaseTestHelper(Jdbi jdbi) {
...
-        ).execute());
+    public static record WebhookDeliveryQueueMessage(int deliveryQueueMessageId,
+                                                     int webhookMessageId,
+                                                     String sentDate,
+                                                     String createdDate,
+                                                     String deliveryResult,
+                                                     int statusCode,
+                                                     DeliveryStatus deliveryStatus,
+                                                     int deliveryCode) {
+        public WebhookDeliveryQueueMessage setDeliveryQueueAndWebhookMessageId(int deliveryQueueMessageId, int webhookMessageId) {
```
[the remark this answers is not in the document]
olatomgds: Done, thank you

PR #1184, opened 2024-02-19
PP-12211: Run CodeQL on Java code only

katstevens:
The CodeQL scans should now only run on changes to Java code (so e.g. it won't run on Docker image Dependabot PRs).

Also removes post-merge workflow trigger - this is probably overkill given we also have weekly scheduled scans.


SandorArpa (approved):
LGTM

PR #1486, opened 2025-02-25
PP-13521 Add total to webhook messages response

james-peacock-gds:
In order to implement better pagination of webhook events in selfservice, we need the total message count to be returned in response to a list webhook messages request.
- Add the total number of messages to the response
- Add unit tests for getting the total message count by webhook id and by status
- Update integration tests


kerrr (approved):
LGTM 🚀

PR #1494, opened 2025-03-25
Add .scot to the allowed domains list

hjvoid:
Updating the domain allow list with `.scot` .


src/main/resources/config/config.yaml
```diff
@@ -103,6 +103,6 @@ webhookMessageSendingQueueProcessorConfig:
 
 # The azurewebsites.net domain was added due to Zendesk ticket #5451010 — it is
 # intended to be temporary and should be gone by the end of August 2023
-liveDataAllowDomains: ["gov.uk", "s118-prod-bsr-acs-common-fa.azurewebsites.net", "cadets.mod.uk"]
+liveDataAllowDomains: ["gov.uk", "s118-prod-bsr-acs-common-fa.azurewebsites.net", "cadets.mod.uk", "scot"]
```
alexbishop1, line 106: I think we want just `edevelopment.scot` — a .[scot domain](https://dot.scot/) can be registered by anyone

alexbishop1 (changes requested):

alexbishop1 (approved):

PR #1503, opened 2025-04-08
PP-13658 refine Dropwizard 4 migration

SandorArpa:
Fine tune Dropwizard 4 bom dependencies.


alexbishop1 (changes requested):
This changes us from using `jakarta.inject-api` version 2.0.1 to version 2.0.1.MR but the differences are likely negligible.

alexbishop1 (approved):

PR #1604, opened 2025-08-13
PP-14198 Change logging message to reflect the attempt number

hjvoid:
Updating the logging wording to clearly denote the time between when the webhook  was scheduled to happen and when it actually happened. The current wording:
`time_to_send_in_millis`
is unclear and there will be another PR to remove it after manually checking this change is searchable in Splunk and changes have been made in the `cyber-security-splunk-apps` repo.


SandorArpa (approved):
LGTM

PR #1860, opened 2026-04-08
PP-15018 add more comprehensive logging for sending webhook messages

AndrzejProchyra:
* Catches, logs and re-throws `Throwable` on webhook message send attempt
* Adds an initial narrow integration test for `SendAttempter` that describes the basic happy-path behaviour and fault behaviour against a mock target.

From studying the logs we see the following behaviour for these situations:

A send attempt is made (retry count is 0)

Nothing else relating to the webhook is logged

Exactly one day later (plus ε) another attempt is made, and generally succeeds. The retry count is also 0

It is clear that on the initial send attempt, we do not get past this line of SendAttempter:

[var response = webhookMessageSender.sendWebhookMessage(queueItem.getWebhookMessageEntity());](https://github.com/alphagov/pay-webhooks/blob/c427d94bd84929c55f878cc2af0f0585e30f899d/src/main/java/uk/gov/pay/webhooks/deliveryqueue/managed/SendAttempter.java#L91)

If an exception is thrown here, it must be caught, because the try-catch block includes Exception as the final catch. Therefore, it is possible that an Error is being thrown by the JVM. We do not log on Errors. It is possible that this Error interferes with the @UnitOfWork in which it is contained, the ScheduledExecutorService worker thread in which it runs, and/or the method by which we obtain the next entry to send (we obtain a “pessimistic write” lock on the entry with the lock timeout set to “skip locked” - see [WebhookDeliveryQueueDao#nextToSend](https://github.com/alphagov/pay-webhooks/blob/c427d94bd84929c55f878cc2af0f0585e30f899d/src/main/java/uk/gov/pay/webhooks/deliveryqueue/dao/WebhookDeliveryQueueDao.java#L35)). Perhaps the entry becomes “visible” again after exactly one day, by some mechanism.

In any case, additionally logging Throwables may very well point us to the source of the problem.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -225,6 +225,7 @@
             <groupId>io.github.netmikey.logunit</groupId>
             <artifactId>logunit-logback</artifactId>
             <version>2.0.0</version>
+            <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>io.rest-assured</groupId>
@@ -288,8 +289,10 @@
                     <outputFile>target/${project.artifactId}-${project.version}-allinone.jar</outputFile>
                     <createDependencyReducedPom>false</createDependencyReducedPom>
                     <transformers>
-                        <transformer implementation="org.apache.maven.plugins.shade.resource.ServicesResourceTransformer"/>
-                        <transformer implementation="org.apache.maven.plugins.shade.resource.ManifestResourceTransformer">
+                        <transformer
+                                implementation="org.apache.maven.plugins.shade.resource.ServicesResourceTransformer"/>
+                        <transformer
+                                implementation="org.apache.maven.plugins.shade.resource.ManifestResourceTransformer">
                             <mainClass>${mainClass}</mainClass>
                         </transformer>
                     </transformers>
--- a/pom.xml
+++ b/pom.xml
@@ -204,10 +204,6 @@
             <scope>test</scope>
         </dependency>
         <dependency>
-            <groupId>org.assertj</groupId>
-            <artifactId>assertj-core</artifactId>
-        </dependency>
-        <dependency>
             <groupId>org.mockito</groupId>
             <artifactId>mockito-junit-jupiter</artifactId>
             <scope>test</scope>
--- a/pom.xml
+++ b/pom.xml
@@ -204,6 +204,10 @@
             <scope>test</scope>
         </dependency>
         <dependency>
+            <groupId>org.assertj</groupId>
+            <artifactId>assertj-core</artifactId>
+        </dependency>
+        <dependency>
             <groupId>org.mockito</groupId>
             <artifactId>mockito-junit-jupiter</artifactId>
             <scope>test</scope>
@@ -220,6 +224,11 @@
             <artifactId>hamcrest-core</artifactId>
             <version>3.0</version>
             <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>io.github.netmikey.logunit</groupId>
+            <artifactId>logunit-logback</artifactId>
+            <version>2.0.0</version>
         </dependency>
         <dependency>
             <groupId>io.rest-assured</groupId>
--- a/src/main/java/uk/gov/pay/webhooks/deliveryqueue/managed/SendAttempter.java
+++ b/src/main/java/uk/gov/pay/webhooks/deliveryqueue/managed/SendAttempter.java
@@ -2,6 +2,8 @@ package uk.gov.pay.webhooks.deliveryqueue.managed;
 
 import com.codahale.metrics.MetricRegistry;
 import io.dropwizard.core.setup.Environment;
+import jakarta.inject.Inject;
+import jakarta.ws.rs.core.Response;
 import net.logstash.logback.marker.LogstashMarker;
 import net.logstash.logback.marker.Markers;
 import org.apache.http.NoHttpResponseException;
@@ -15,8 +17,6 @@ import uk.gov.pay.webhooks.deliveryqueue.dao.WebhookDeliveryQueueEntity;
 import uk.gov.pay.webhooks.message.WebhookMessageSender;
 import uk.gov.pay.webhooks.validations.CallbackUrlDomainNotOnAllowListException;
 
-import jakarta.inject.Inject;
-import jakarta.ws.rs.core.Response;
 import java.io.IOException;
 import java.net.MalformedURLException;
 import java.net.SocketTimeoutException;
@@ -72,7 +72,7 @@ public class SendAttempter {
         Instant start = instantSource.instant();
 
         URL url = null;
-        
+
         try {
             url = new URL(webhook.getCallbackUrl().strip());
         } catch (MalformedURLException e) {
@@ -87,7 +87,7 @@ public class SendAttempter {
                             .and(Markers.append(WEBHOOK_CALLBACK_URL_DOMAIN, url.getHost()))
                             .and(Markers.append(WEBHOOK_MESSAGE_DELAY_BETWEEN_ATTEMPT_SCHEDULED_SEND_AT_TIME_AND_NOW, Duration.between(queueItem.getSendAt(), instantSource.instant()).toMillis())),
                     "Sending webhook message started"
-            ); 
+            );
             var response = webhookMessageSender.sendWebhookMessage(queueItem.getWebhookMessageEntity());
 
             var statusCode = response.getStatusLine().getStatusCode();
@@ -116,6 +116,11 @@ public class SendAttempter {
             handleResponse(queueItem, DeliveryStatus.WILL_NOT_SEND, null, "Violates security rules", retryCount, start);
         } catch (Exception e) {
             handleGenericException(queueItem, retryCount, start, e);
+        } catch (Throwable throwable) {
+            LOGGER.atError()
+                    .setCause(throwable)
+                    .log("Error during webhook message send");
+            throw throwable;
         }
     }
 
@@ -138,25 +143,25 @@ public class SendAttempter {
         handleResponse(webhookDeliveryQueueEntity, status, statusCode, reason, retryCount, startTime, Optional.empty());
     }
 
-    private void handleResponse(WebhookDeliveryQueueEntity webhookDeliveryQueueEntity, 
-                                DeliveryStatus status, 
-                                Integer statusCode, 
-                                String reason, 
-                                Long retryCount, 
+    private void handleResponse(WebhookDeliveryQueueEntity webhookDeliveryQueueEntity,
+                                DeliveryStatus status,
+                                Integer statusCode,
+                                String reason,
+                                Long retryCount,
                                 Instant startTime,
                                 Optional<URL> domain) {
         var responseTime = Duration.between(startTime, instantSource.instant());
-        
+
         LogstashMarker logstashMarker = Markers.append(HTTP_STATUS, statusCode)
                 .and(Markers.append(WEBHOOK_MESSAGE_RETRY_COUNT, retryCount))
                 .and(Markers.append(STATE_TRANSITION_TO_STATE, status))
                 .and(Markers.append(RESPONSE_TIME, responseTime.toMillis()))
                 .and(Markers.append(WEBHOOK_MESSAGE_ATTEMPT_RESPONSE_REASON, reason));
-        
+
         domain.ifPresent(d -> logstashMarker.and(Markers.append(WEBHOOK_CALLBACK_URL_DOMAIN, d.getHost())));
-        
-        LOGGER.info(logstashMarker, "Sending webhook message finished"); 
-        
+
+        LOGGER.info(logstashMarker, "Sending webhook message finished");
+
         webhookDeliveryQueueDao.recordResult(webhookDeliveryQueueEntity, reason, responseTime, statusCode, status, metricRegistry);
         webhookDeliveryQueueEntity.getWebhookMessageEntity().setLastDeliveryStatus(status);
 
@@ -184,7 +189,7 @@ public class SendAttempter {
             default -> null;
         };
     }
-    
+
     private String getReasonFromStatusCode(int statusCode) {
         return Stream.of(String.valueOf(statusCode), Response.Status.fromStatusCode(statusCode).getReasonPhrase())
                 .filter(Objects::nonNull)
--- a/src/test/java/uk/gov/pay/webhooks/deliveryqueue/managed/SendAttempterIT.java
+++ b/src/test/java/uk/gov/pay/webhooks/deliveryqueue/managed/SendAttempterIT.java
@@ -110,7 +110,7 @@ class SendAttempterIT {
     }
 
     @Test
-    void should_log_connection_reset_by_peer_fault_with_error_message() {
+    void should_log_error_message_when_connection_is_reset_by_peer() {
         givenThat(post(CALLBACK_URL)
                 .willReturn(aResponse()
                         .withFault(CONNECTION_RESET_BY_PEER)));
--- a/src/test/java/uk/gov/pay/webhooks/deliveryqueue/managed/SendAttempterIT.java
+++ b/src/test/java/uk/gov/pay/webhooks/deliveryqueue/managed/SendAttempterIT.java
@@ -125,7 +125,7 @@ class SendAttempterIT {
     }
 
     @Test
-    void should_log_random_data_then_close_fault() {
+    void should_log_when_response_is_random_data_and_the_connection_is_closed() {
         givenThat(post(CALLBACK_URL)
                 .willReturn(aResponse()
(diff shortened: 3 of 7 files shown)


pom.xml
```diff
@@ -203,6 +203,10 @@
             <artifactId>junit-vintage-engine</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.assertj</groupId>
```
kbottla: Can we stick to Hamcrest until there is a discussion about AssertJ with the team and we have a clear direction? Introducing as we go could confuse the team, and we'll need to plan some upskilling too.

src/main/java/uk/gov/pay/webhooks/deliveryqueue/managed/SendAttempter.java
```diff
@@ -116,6 +116,11 @@ public void attemptSend(WebhookDeliveryQueueEntity queueItem) {
             handleResponse(queueItem, DeliveryStatus.WILL_NOT_SEND, null, "Violates security rules", retryCount, start);
         } catch (Exception e) {
             handleGenericException(queueItem, retryCount, start, e);
+        } catch (Throwable throwable) {
+            LOGGER.atError()
+                    .setCause(throwable)
+                    .log("Error during webhook message send");
+            throw throwable;
```
DomBelcher, line 123: What would currently happen with any uncaught `Throwable` or `Error`? would it just crash the app, or is there somewhere it could be caught?

AndrzejProchyra: That is an excellent question, @DomBelcher. As I mention in the PR description, the `Error` would escape the containing `@UnitOfWork`, and perhaps any cleanup behaviour is prevented from happening (such as releasing the lock on the record). The [documentation for `ScheduledExecutorService`](https://docs.oracle.com/en/java/javase/25/docs/api/java.base/java/util/concurrent/ScheduledExecutorService.html#scheduleWithFixedDelay(java.lang.Runnable,long,long,java.util.concurrent.TimeUnit)) implies that the task will stop executing.

This is a worker thread, and not the main thread, so I wouldn't expect it to crash the app, and we are not seeing restarts neare the time of the bug that we are seeing.

src/test/java/uk/gov/pay/webhooks/deliveryqueue/managed/SendAttempterIT.java
```diff
@@ -0,0 +1,210 @@
...
+
+@WireMockTest
+@ExtendWith(MockitoExtension.class)
+class SendAttempterIT {
+
+    @RegisterExtension
+    LogCapturer logs = LogCapturer.create().captureForLogger(ROOT_LOGGER_NAME);
+
+    @Mock
+    private WebhookDeliveryQueueDao webhookDeliveryQueueDaoSpy;
```
kbottla: This could be mockSomething as it is not a [spy](https://site.mockito.org/javadoc/current/org/mockito/Spy.html)

```suggestion
    private WebhookDeliveryQueueDao mockWebhookDeliveryQueueDao;
```

src/test/java/uk/gov/pay/webhooks/deliveryqueue/managed/SendAttempterIT.java
```diff
@@ -0,0 +1,210 @@
...
+    private WebhookDeliveryQueueEntity enqueuedItem;
+
+    @BeforeEach
+    void setUp(WireMockRuntimeInfo wmRuntimeInfo) {
+        sendAttempter = buildSendAttempter();
+        enqueuedItem = buildWebhookDeliveryQueueEntityWithCallbackURL(wmRuntimeInfo.getHttpBaseUrl() + CALLBACK_URL);
+    }
+
+    @Nested
+    class when_send_successful {
```
kbottla, line 76: Can the class name be as per standard convention? 

```suggestion
    class whenSendSuccessful {
```

src/test/java/uk/gov/pay/webhooks/deliveryqueue/managed/SendAttempterIT.java
```diff
@@ -0,0 +1,210 @@
...
+        sendAttempter.attemptSend(enqueuedItem);
+
+        var loggingEvent = logs.assertContains("Exception caught by request");
+        var markers = loggingEvent.getMarkers();
+        assertThat(markers.size(), is(1));
+        assertThat(markers.getFirst(), hasToString("error_message=Connection reset"));
+    }
+
+    @Test
+    void should_log_random_data_then_close_fault() {
```
kbottla: We could describe the test more clearly, so it is easy to understand?


```suggestion
    void should_log_when_response_is_random_data_and_the_connection_is_closed() {
```

src/test/java/uk/gov/pay/webhooks/deliveryqueue/managed/SendAttempterIT.java
```diff
@@ -0,0 +1,210 @@
...
+        sendAttempter.attemptSend(enqueuedItem);
+
+        var loggingEvent = logs.assertContains("Exception caught by request");
+        var markers = loggingEvent.getMarkers();
+        assertThat(markers.size(), is(1));
+        assertThat(markers.getFirst(), hasToString("error_message=null"));
+    }
+
+    @Test
+    void should_log_empty_response_fault() {
```
kbottla:
```suggestion
    void should_log_empty_error_response() {
```

src/test/java/uk/gov/pay/webhooks/deliveryqueue/managed/SendAttempterIT.java
```diff
@@ -0,0 +1,210 @@
...
+        givenThat(post(CALLBACK_URL)
+                .willReturn(aResponse()
+                        .withFault(EMPTY_RESPONSE)));
+
+        sendAttempter.attemptSend(enqueuedItem);
+
+        logs.assertContains("Request timed out"); // Is this the right thing?
```
kbottla: Can the comment be removed?

Doesn't sound like we should log `Request timed out` for NoHttpResponseException, though. But it can be addressed in a separate PR.

src/test/java/uk/gov/pay/webhooks/deliveryqueue/managed/SendAttempterIT.java
```diff
@@ -0,0 +1,210 @@
...
+                webhookMessageSender,
+                environment
+        );
+    }
+
+    private static @NonNull WebhookMessageSender buildWebhookMessageSender() {
+        var httpClient = HttpClients.createDefault();
+        var httpPostFactory = new HttpPostFactory();
+        var objectMapper = new ObjectMapper();
+        var callbackUrlService = mock(CallbackUrlService.class); // Using Mockito since the WebhooksConfig has no constructor
```
kbottla: Can the comment be removed? It is clear that we are mocking the class

src/test/java/uk/gov/pay/webhooks/deliveryqueue/managed/SendAttempterIT.java
```diff
@@ -0,0 +1,210 @@
...
+                .buld();
+        var webhookMessage = aWebhookMessageEntity()
+                .withWebhookEntity(webhook)
+                .build();
+        return aWebhookDeliveryQueueEntity()
+                .withWebhookMessageEntity(webhookMessage)
+                .build();
+    }
+
+    private @NonNull SendAttempter buildSendAttempter() {
```
kbottla: Any advantages of using ` @NonNull` annotation? It is helpful for method parameters or some reused methods across many tests. It could become noise if we start using it everywhere.

src/test/java/uk/gov/pay/webhooks/deliveryqueue/managed/SendAttempterIT.java
```diff
@@ -0,0 +1,210 @@
+package uk.gov.pay.webhooks.deliveryqueue.managed;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.github.tomakehurst.wiremock.http.Fault;
+import com.github.tomakehurst.wiremock.junit5.WireMockRuntimeInfo;
+import com.github.tomakehurst.wiremock.junit5.WireMockTest;
+import io.dropwizard.core.setup.Environment;
+import io.github.netmikey.logunit.api.LogCapturer;
+import org.apache.http.impl.client.HttpClients;
+import org.jspecify.annotations.NonNull;
```
kbottla: It looks like jspecify is not a direct dependency. It would be good to avoid using it in our tests.

pom.xml
```diff
@@ -221,6 +221,11 @@
             <version>3.0</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>io.github.netmikey.logunit</groupId>
```
kbottla, line 225: The repo doesn't look very active, with only one contributor. Is this recommended somewhere?

src/test/java/uk/gov/pay/webhooks/deliveryqueue/managed/SendAttempterIT.java
AndrzejProchyra: Not a Mockito `Spy`, no, but I was applying the Meszaros [Test Double Patterns](http://xunitpatterns.com/Test%20Double%20Patterns.html) language. That is to say:

> Test Spy
> Use a Test Double to capture the indirect output calls made to another component by the SUT for later verification by the test.

as opposed to 

> Mock Object
> Replace an object the system under test (SUT) depends on with a test-specific object that verifies it is being used correctly by the SUT.

In the case of the `WebhookDeliveryQueueDao` we simply verify that `enqueueFrom()` was called with `DeliveryStatus.PENDING` and an `Instant` that is after the `Instant` that we consider to be "now". We do not stub a return value.

src/test/java/uk/gov/pay/webhooks/deliveryqueue/managed/SendAttempterIT.java
AndrzejProchyra: The reason that I have written it like this is because this class is being used as a context for hierarchical tests, rather than something to be called by our code. If we were to [set the default display name generator](https://docs.junit.org/6.0.3/writing-tests/display-names.html#generator-default) to `ReplaceUnderscores` we would have nice test reports that look like this:
image

src/test/java/uk/gov/pay/webhooks/deliveryqueue/managed/SendAttempterIT.java
AndrzejProchyra: That is an excellent improvement. Thank you.

src/test/java/uk/gov/pay/webhooks/deliveryqueue/managed/SendAttempterIT.java
AndrzejProchyra: It is true that the test name is confusing, like the one above, but we are not logging an empty error response here, we are logging when we receive [a completely empty response](https://wiremock.org/docs/simulating-faults/#bad-responses)... Perhaps just `should_log_empty_response`?

src/test/java/uk/gov/pay/webhooks/deliveryqueue/managed/SendAttempterIT.java
AndrzejProchyra: Yes, that is a good idea

src/test/java/uk/gov/pay/webhooks/deliveryqueue/managed/SendAttempterIT.java
AndrzejProchyra: Yes, OK. I added the comment because I would have used the real `CallbackUrlService` except that many of our `Config` classes do not have setters or constructors and rely on reflection for deserialisation. Perhaps it would be helpful to add setters and/or constructors for use in focused tests.

src/test/java/uk/gov/pay/webhooks/deliveryqueue/managed/SendAttempterIT.java
AndrzejProchyra: Yes, this is true. This annotation was added automatically by IntelliJ when I extracted the method. I am not sure why IntelliJ has started doing this. I shall attempt to configure my IntelliJ to stop doing this.

src/test/java/uk/gov/pay/webhooks/deliveryqueue/managed/SendAttempterIT.java
AndrzejProchyra: See above. This was added by IntelliJ's 'Extract Method' refactoring and I have removed it.

pom.xml
AndrzejProchyra: As mentioned in my Firebreak Week presentation I implemented a custom JUnit extension to simplify testing logging behaviour, but later saw that LogUnit was listed as a [third-party extension in the JUnit wiki](https://github.com/junit-team/junit-framework/wiki/Third-party-Extensions#junit-jupiter-extensions). I considered that it would be better to use such a "community maintained" dependency in preference to maintaining one ourselves. Perhaps we could reconsider this?

src/test/java/uk/gov/pay/webhooks/deliveryqueue/managed/SendAttempterIT.java
kbottla: It won't be clear why it is called "spy" when it's created by `@Mock`, which still creates a mock object even when tests don't mock method calls. This causes confusion about its intention.

src/test/java/uk/gov/pay/webhooks/deliveryqueue/managed/SendAttempterIT.java
```diff
@@ -0,0 +1,209 @@
...
+
+            assertThat(logs.size(), is(2));
+            logs.assertContains("Sending webhook message started");
+            logs.assertContains("Sending webhook message finished");
+        }
+    }
+
+    @Test
+    void should_log_connection_reset_by_peer_fault_with_error_message() {
+        givenThat(post(CALLBACK_URL)
```
kbottla, line 114: Can `fault` be removed from test names? It is introducing something from Wiremock, rather than using the application terminology?

```suggestion
    void should_log_error_message_when_connection_is_reset_by_peer() {
```

src/test/java/uk/gov/pay/webhooks/deliveryqueue/managed/SendAttempterIT.java
kbottla: Maybe it can be something like below

` void should_log_when_empty_error_response_is_received() {`

pom.xml
kbottla: It doesn't look like doing much. I am fine to use it or go our own way. 

Can the scope be `test` for this library?

src/test/java/uk/gov/pay/webhooks/deliveryqueue/managed/SendAttempterIT.java
AndrzejProchyra: The intent here was to be intentional about the type of test double I was using. If this is, instead, causing confusion it ,of course, makes sense to use the Mockito terminology. I have renamed this, as you suggested.

src/test/java/uk/gov/pay/webhooks/deliveryqueue/managed/SendAttempterIT.java
AndrzejProchyra: Yes, that makes sense

pom.xml
AndrzejProchyra: Yes, of course. This was an omission on my part

src/test/java/uk/gov/pay/webhooks/deliveryqueue/managed/SendAttempterIT.java
AndrzejProchyra: The trouble is that there is a distinction between non-200 HTTP "error" responses and "faults" in the underlying connection, where the application layer protocol has broken down. I am wondering what a good name would be for the parameterised test `should_enqueue_for_retry_on_fault(Fault fault)`, because retries on "error" responses are a separate behaviour.

kbottla: Could it be `should_enqueue_for_retry_on_connector_error(Fault connectionError)`?

AndrzejProchyra: OK, I have changed the "fault" language to "connection error" although, of course, the WireMock code still calls it a "fault" in the stubbing DSL.

kbottla (approved):
LGTM! Please squash commits before merging

PR #1973, opened 2026-07-20
PP-15585 refactor to use apache5 client

SandorArpa:
Refactor to use Apache v5 client.

Refactor to remove Apache v4 dependencies from the code.

Refactor AWS SqsClient to use Apache5 also.

Refactor tests.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -2,7 +2,7 @@
 <project
         xmlns="http://maven.apache.org/POM/4.0.0"
         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-        xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+        xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
 
     <modelVersion>4.0.0</modelVersion>
 
@@ -116,6 +116,10 @@
             <artifactId>sqs</artifactId>
         </dependency>
         <dependency>
+            <groupId>software.amazon.awssdk</groupId>
+            <artifactId>apache5-client</artifactId>
+        </dependency>
+        <dependency>
             <groupId>io.dropwizard.metrics</groupId>
             <artifactId>metrics-core</artifactId>
         </dependency>
@@ -126,6 +130,10 @@
         <dependency>
             <groupId>jakarta.inject</groupId>
             <artifactId>jakarta.inject-api</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.httpcomponents.client5</groupId>
+            <artifactId>httpclient5</artifactId>
         </dependency>
 
         <!-- Main dependencies that need explicit versions -->
@@ -169,12 +177,6 @@
             <artifactId>logging-dropwizard-5</artifactId>
             <version>${pay-java-commons.version}</version>
         </dependency>
-        <dependency>
-            <groupId>org.apache.httpcomponents</groupId>
-            <artifactId>httpclient</artifactId>
-            <version>4.5.14</version>
-        </dependency>
-
 
         <!-- Test dependencies that are imported from one of the BOMs specified
              in <dependencyManagement> so no explicit versions needed -->
--- a/src/main/java/uk/gov/pay/webhooks/app/WebhooksModule.java
+++ b/src/main/java/uk/gov/pay/webhooks/app/WebhooksModule.java
@@ -7,16 +7,20 @@ import io.dropwizard.hibernate.HibernateBundle;
 import io.dropwizard.hibernate.UnitOfWorkAwareProxyFactory;
 import jakarta.inject.Singleton;
 import jakarta.ws.rs.client.Client;
-import org.apache.http.client.config.CookieSpecs;
-import org.apache.http.client.config.RequestConfig;
-import org.apache.http.conn.ssl.SSLConnectionSocketFactory;
-import org.apache.http.impl.client.CloseableHttpClient;
-import org.apache.http.impl.client.HttpClientBuilder;
-import org.apache.http.impl.conn.PoolingHttpClientConnectionManager;
-import org.apache.http.ssl.SSLContexts;
+import org.apache.hc.client5.http.config.RequestConfig;
+import org.apache.hc.client5.http.cookie.StandardCookieSpec;
+import org.apache.hc.client5.http.impl.classic.CloseableHttpClient;
+import org.apache.hc.client5.http.impl.classic.HttpClientBuilder;
+import org.apache.hc.client5.http.impl.io.PoolingHttpClientConnectionManager;
+import org.apache.hc.client5.http.impl.io.PoolingHttpClientConnectionManagerBuilder;
+import org.apache.hc.client5.http.ssl.DefaultHostnameVerifier;
+import org.apache.hc.client5.http.ssl.SSLConnectionSocketFactory;
+import org.apache.hc.core5.http.io.SocketConfig;
+import org.apache.hc.core5.util.Timeout;
 import org.hibernate.SessionFactory;
 import software.amazon.awssdk.auth.credentials.AwsBasicCredentials;
 import software.amazon.awssdk.auth.credentials.StaticCredentialsProvider;
+import software.amazon.awssdk.http.apache5.Apache5HttpClient;
 import software.amazon.awssdk.regions.Region;
 import software.amazon.awssdk.services.sqs.SqsClient;
 import software.amazon.awssdk.services.sqs.SqsClientBuilder;
@@ -24,7 +28,9 @@ import uk.gov.pay.webhooks.message.HttpPostFactory;
 import uk.gov.pay.webhooks.message.WebhookMessageSignatureGenerator;
 import uk.gov.pay.webhooks.util.IdGenerator;
 
+import javax.net.ssl.SSLContext;
 import java.net.URI;
+import java.security.NoSuchAlgorithmException;
 import java.time.InstantSource;
 import java.util.concurrent.TimeUnit;
 
@@ -74,10 +80,27 @@ public class WebhooksModule extends AbstractModule {
     @Provides
     public PoolingHttpClientConnectionManager getConnectionPoolManager() {
         int connectionPoolSize = configuration.getWebhookMessageSendingQueueProcessorConfig().getHttpClientConnectionPoolSize();
-        PoolingHttpClientConnectionManager poolingConnManager
-                = new PoolingHttpClientConnectionManager();
-        poolingConnManager.setMaxTotal(connectionPoolSize);
-        poolingConnManager.setDefaultMaxPerRoute(connectionPoolSize);
+        SSLConnectionSocketFactory sslConnectionSocketFactory;
+        try {
+            sslConnectionSocketFactory = new SSLConnectionSocketFactory(
+                    SSLContext.getDefault(),
+                    new String[]{"TLSv1.2", "TLSv1.3"},
+                    null,
+                    new DefaultHostnameVerifier()
+            );
+        } catch (NoSuchAlgorithmException e) {
+            throw new RuntimeException("Unable to create SSL connection socket factory", e);
+        }
+        
+        PoolingHttpClientConnectionManager poolingConnManager = PoolingHttpClientConnectionManagerBuilder
+                .create()
+                .setSSLSocketFactory(sslConnectionSocketFactory)
+                .setMaxConnTotal(connectionPoolSize)
+                .setMaxConnPerRoute(connectionPoolSize)
+                .setDefaultSocketConfig(SocketConfig.custom()
+                        .setSoTimeout(Timeout.ofMinutes(1))
+                        .build())
+                .build();
 
         return poolingConnManager;
     }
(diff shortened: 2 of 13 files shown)


pom.xml
```diff
@@ -169,12 +173,6 @@
             <artifactId>logging-dropwizard-5</artifactId>
             <version>${pay-java-commons.version}</version>
         </dependency>
-        <dependency>
-            <groupId>org.apache.httpcomponents</groupId>
-            <artifactId>httpclient</artifactId>
-            <version>4.5.14</version>
-        </dependency>
```
AndrzejProchyra, line 176: We should replace this with

```xml
<dependency>                                            
    <groupId>org.apache.httpcomponents.client5</groupId>
    <artifactId>httpclient5</artifactId>                
</dependency>                                           
```

If we remove the direct dependency entirely, we are relying on the fact that it is currently a transitive dependency of `org.apache.httpcomponents.client5:httpclient5`, but we should make our explicit dependencies explicit.

src/test/java/uk/gov/pay/webhooks/message/WebhookMessageSenderTest.java
```diff
@@ -102,7 +101,7 @@ public void setUp() throws JsonProcessingException, InvalidKeyException {
     }
 
     @Test
-    void constructsHttpRequestAndReturnsHttpResponse() throws IOException, InvalidKeyException, InterruptedException {
+    void constructsHttpRequestAndReturnsHttpResponse() throws IOException, InvalidKeyException, InterruptedException, URISyntaxException {
```
AndrzejProchyra, line 104: Could this just be `throws Exception`?

src/test/java/uk/gov/pay/webhooks/deliveryqueue/managed/SendAttempterIT.java
```diff
@@ -131,10 +132,11 @@ void should_log_when_response_is_random_data_and_the_connection_is_closed() {
 
         sendAttempter.attemptSend(enqueuedItem);
 
-        var loggingEvent = logs.assertContains("Exception caught by request");
+        logs.assertContains("Request timed out");
+        var loggingEvent = logs.assertContains("Sending webhook message finished");
         var markers = loggingEvent.getMarkers();
         assertThat(markers.size(), is(1));
-        assertThat(markers.getFirst(), hasToString("error_message=null"));
+        assertThat(markers.getFirst().toString(), containsString("reason=HTTP Timeout"));
```
AndrzejProchyra, line 139: This could also be

`assertThat(markers.getFirst(), hasToString(containsString("reason=HTTP Timeout")))`

Which may be more expressive, since we are asserting on a `Marker`.

src/main/java/uk/gov/pay/webhooks/app/WebhooksModule.java
```diff
@@ -74,8 +80,25 @@ public WebhookMessageDeletionConfig webhookMessageDeletionConfig() {
...
+                    null,
+                    new DefaultHostnameVerifier()
+            );
+        } catch (NoSuchAlgorithmException e) {
+            throw new RuntimeException("Unable to create SSL connection socket factory", e);
+        }
+        
+        PoolingHttpClientConnectionManager poolingConnManager = PoolingHttpClientConnectionManagerBuilder
+                .create()
+                .setSSLSocketFactory(sslConnectionSocketFactory)
```
AndrzejProchyra, line 97: This is deprecated in favour of `setTlsSocketStrategy`. Perhaps we can do something like this:

```java
.setTlsSocketStrategy(ClientTlsStrategyBuilder.create()
                        .setSslContext(SSLContexts.createSystemDefault())
                        .setTlsVersions(TLS.V_1_2, TLS.V_1_3)
                        .buildClassic())
```

src/main/java/uk/gov/pay/webhooks/app/WebhooksModule.java
```diff
@@ -87,25 +110,17 @@ public PoolingHttpClientConnectionManager getConnectionPoolManager() {
     public CloseableHttpClient httpClient(PoolingHttpClientConnectionManager poolingConnManager) {
         var timeoutInMillis = Math.toIntExact(configuration.getWebhookMessageSendingQueueProcessorConfig().getRequestTimeout().toMilliseconds());
         var config = RequestConfig.custom()
-                .setConnectTimeout(timeoutInMillis)
-                .setConnectionRequestTimeout(timeoutInMillis)
-                .setSocketTimeout(timeoutInMillis)
-                .setCookieSpec(CookieSpecs.STANDARD)
+                .setConnectTimeout(Timeout.of(timeoutInMillis, TimeUnit.MILLISECONDS))
```
AndrzejProchyra, line 113: This is deprecated in favour of setting a timeout on a connection. Instead we could make the following call to the `PoolingHttpClientConnectionManagerBuilder`:

```java
.setDefaultConnectionConfig(ConnectionConfig.custom()
                        .setConnectTimeout(Timeout.of(timeoutInMillis, TimeUnit.MILLISECONDS))
                        .build())
```

src/main/java/uk/gov/pay/webhooks/app/WebhooksModule.java
```diff
@@ -74,8 +80,25 @@ public WebhookMessageDeletionConfig webhookMessageDeletionConfig() {
...
+        
+        PoolingHttpClientConnectionManager poolingConnManager = PoolingHttpClientConnectionManagerBuilder
+                .create()
+                .setSSLSocketFactory(sslConnectionSocketFactory)
+                .setDefaultSocketConfig(SocketConfig.custom()
+                        .setSoTimeout(Timeout.ofMinutes(1))
+                        .build())
+                .build();
         poolingConnManager.setMaxTotal(connectionPoolSize);
         poolingConnManager.setDefaultMaxPerRoute(connectionPoolSize);
```
AndrzejProchyra: We could chain these onto the calls to `PoolingHttpClientConnectionManagerBuilder` above:

```java
.setMaxConnTotal(connectionPoolSize)   
.setMaxConnPerRoute(connectionPoolSize)
```

and then just return the result of the whole builder chain.

pom.xml
SandorArpa: Yes, same as any other Java projects that depends on httpclient5. It's a transative dependency from Dropwizard.

src/test/java/uk/gov/pay/webhooks/message/WebhookMessageSenderTest.java
SandorArpa: No, we don't do that. It's a convention we use in all Java projects. Same as we don't use import java.*

src/main/java/uk/gov/pay/webhooks/app/WebhooksModule.java
SandorArpa: Well, yet we use it.

src/main/java/uk/gov/pay/webhooks/app/WebhooksModule.java
SandorArpa: we could but we keep it as is for readability

src/main/java/uk/gov/pay/webhooks/app/WebhooksModule.java
SandorArpa: Yes, I agree we could. However **all** timeouts are set on RequestConfig.custom() so it's better to keep it together.

src/main/java/uk/gov/pay/webhooks/app/WebhooksModule.java
SandorArpa: This story is about replacing like to like as closely as possible. We can create another story which refactors all deprecations.

src/test/java/uk/gov/pay/webhooks/deliveryqueue/managed/SendAttempterIT.java
SandorArpa: why is `hasToString(containsString())` better than simple `constainsString()`?

pom.xml
AndrzejProchyra: Yes, but we have an implicit **direct** dependency on this artifact. We import its classes directly in **8** of our own classes. We should therefore declare it to be an explicit dependency of our code, rather than an implementation detail of some other direct dependency. Because that could change on any update and break our code.

`mvn dependency:tree -Dincludes="org.apache.httpcomponents.client5:httpclient5"`

Gives the following output

```
[INFO] Scanning for projects...
[INFO] 
[INFO] ----------------------< uk.gov.pay:pay-webhooks >-----------------------
[INFO] Building pay-webhooks 0.1-SNAPSHOT
[INFO]   from pom.xml
[INFO] --------------------------------[ jar ]---------------------------------
[INFO] 
[INFO] --- dependency:3.7.0:tree (default-cli) @ pay-webhooks ---
[INFO] uk.gov.pay:pay-webhooks:jar:0.1-SNAPSHOT
[INFO] \- software.amazon.awssdk:apache5-client:jar:2.53.0:compile
[INFO]    \- org.apache.httpcomponents.client5:httpclient5:jar:5.6.1:compile
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  2.057 s
[INFO] Finished at: 2026-08-18T15:17:06+01:00
[INFO] ------------------------------------------------------------------------
```

We only have these classes on our classpath because it happens to be an implementation detail of our declared dependency `software.amazon.awssdk:apache5-client`.

src/test/java/uk/gov/pay/webhooks/message/WebhookMessageSenderTest.java
AndrzejProchyra: Can you explain why? It seems like unnecessary noise to me. The purpose of declared exceptions (checked or runtime) is so that the caller knows that it must deal with them. Here the caller is the test framework, and it does not care what the specific type of an exception is. A test fails if any exception is thrown.

src/test/java/uk/gov/pay/webhooks/deliveryqueue/managed/SendAttempterIT.java
AndrzejProchyra: Just because it seems to be more intention revealing.

The current form says: "Assert that the value returned by calling `toString()` on the first `Marker` contains the following string". When it fails we get:

```
java.lang.AssertionError: 
Expected: a string containing "SOMETHING UNEXPECTED"
     but: was "status_code=null, retry_count=0, to_state=FAILED, response_time=0, reason=HTTP Timeout, domain=localhost"
Expected :a string containing "SOMETHING UNEXPECTED"
Actual   :"status_code=null, retry_count=0, to_state=FAILED, response_time=0, reason=HTTP Timeout, domain=localhost"
```

The other form says: "Assert that the first `Marker`'s "toString" containing the following". When it fails we get:

```
java.lang.AssertionError: 
Expected: with toString() a string containing "SOMETHING UNEXPECTED"
     but: toString() was "status_code=null, retry_count=0, to_state=FAILED, response_time=0, reason=HTTP Timeout, domain=localhost"
```

Perhaps the difference is subtle, but the thing that we are asserting on is a property of a `Marker` (what does its "toString" contain?) rather than on a `String` that we happened to extract from someplace.

src/test/java/uk/gov/pay/webhooks/message/WebhookMessageSenderTest.java
SandorArpa: I don't know why. I just follow the conventions and I think we should not to change one and not all others.

src/test/java/uk/gov/pay/webhooks/deliveryqueue/managed/SendAttempterIT.java
SandorArpa: Readability and simplicity. Adding a nested containsString() makes no sense. Either we check all hasToString or part of it. Also, here we are not testing whether it's a marker or not or it's behaving like a marker or not. We're testing whether it contains a certain string. Therefore containsString() is adequate enough, without nesting it.

src/main/java/uk/gov/pay/webhooks/app/WebhooksModule.java
AndrzejProchyra: Yes, because the call needs to be made to `PoolingHttpClientConnectionManagerBuilder`, so something like

```java
PoolingHttpClientConnectionManager poolingConnManager = PoolingHttpClientConnectionManagerBuilder.create()
        .setTlsSocketStrategy(ClientTlsStrategyBuilder.create()                                           
                .setSslContext(SSLContexts.createSystemDefault())                                         
                .setTlsVersions(TLS.V_1_2, TLS.V_1_3)                                                     
                .buildClassic())                                                                          
        .setDefaultSocketConfig(SocketConfig.custom()                                                     
                .setSoTimeout(Timeout.ofMinutes(1))                                                       
                .build())                                                                                 
        .setDefaultConnectionConfig(ConnectionConfig.custom()                                             
                .setConnectTimeout(Timeout.of(timeoutInMillis, TimeUnit.MILLISECONDS))                    
                .build())                                                                                 
        .setMaxTotal(connectionPoolSize)                                                                  
        .setDefaultMaxPerRoute(connectionPoolSize)                                                        
        .build();                                                                                         
```

src/main/java/uk/gov/pay/webhooks/app/WebhooksModule.java
AndrzejProchyra: I don't follow. We:

1. Create a `PoolingHttpClientConnectionManagerBuilder`
2. Set some properties on it
3. Ask the builder to build a `PoolingHttpClientConnectionManager`
4. Set a couple of things on the newly built connection manager

Why not merge 4 into 3 and just set everything we need before building as a final step?

SandorArpa: The idea is that PoolingHttpClientConnectionManager can be used for something else as well. See IdleConnectionMonitor where it's injected. IdleConnectionMonitor does not need what we set in step 4. Steps 4 are there for the CloseableHttpClient, not the PoolingHttpClientConnectionManager.

src/test/java/uk/gov/pay/webhooks/deliveryqueue/managed/SendAttempterIT.java
AndrzejProchyra: If you prefer. My feeling is still that we are matching on some property of a `Marker` ("something" about its string representation), rather than on a property of some particular string. Hamcrest intentionally provides two forms of the [hasToString](https://javadoc.io/static/org.hamcrest/hamcrest/3.0/org/hamcrest/Matchers.html#hasToString-org.hamcrest.Matcher-) matcher. One that takes an expected "toString" and one that takes a matcher for inexact matches. It is unfortunate that we are needing to match on a "toString" in the first place. Perhaps there is a better way.

src/main/java/uk/gov/pay/webhooks/app/WebhooksModule.java
AndrzejProchyra: > Yes, I agree we could. However **all** timeouts are set on RequestConfig.custom() so it's better to keep it together.

Yes, OK, but this form has been deprecated and could be removed on an update.

SandorArpa: It might be removed in a future version, I agree. But refactoring because of potential removal is not part of this story.

src/main/java/uk/gov/pay/webhooks/app/WebhooksModule.java
AndrzejProchyra: I'm afraid that I don't understand. As far as I can tell Guice instantiates exactly one instance of `PoolingHttpClientConnectionManager`, with all the properties set, whether by using the builder or setting them afterwards in the scope of the `@Provides` method. The same instance is injected in `IdleConnectionMonitor` and the `httpClient` `@Provides` method, right?

SandorArpa: Well, perhaps. Anyway, I don't think we should set half of the timeout in one place and the other ones in another place, based on that one of them might be removed in the future.

src/main/java/uk/gov/pay/webhooks/app/WebhooksModule.java
AndrzejProchyra: I have created a bare-bones ticket: https://payments-platform.atlassian.net/browse/PP-15772

SandorArpa: Nice, thanks

AndrzejProchyra (dismissed):

AndrzejProchyra (approved):

